PR #1, opened 2026-06-30
perf: tune batch and MTP for working DSpark speculation

0rand:
Bump MAX_NUM_BATCHED_TOKENS from 8192 to 16384 — MTP=5 with batch=8K produced only the first useful speculative token; remaining tokens were garbage (<50% by token #2).

Drop default MTP from 5 to 3 with batch=16K — 70-80% correct across all speculative tokens (validated on 1M context, 2x DGX Spark TP=2).

Bump GPU_MEMORY_UTILIZATION from 0.80 to 0.84 to reclaim KV cache consumed by the larger batch budget (production validated at 0.87 for 2M KV cache; 0.84 is the safe default).

Parameterize --port via VLLM_PORT env var (was hardcoded 8888).

Wire num_speculative_tokens in speculative-config to MTP_NUM_TOKENS env var via docker-compose 48006 escape, preventing silent config drift between .env and runtime.


tonyd2wild:
Thanks @0rand — this PR aged remarkably well, and we've now dispositioned every piece of it on the rig (2× DGX Spark TP=2, current main). The full ledger:

**Adopted with credit (d20efb0):**
- `--port ${VLLM_PORT:-8888}` — merged as a maintainer commit with your co-authorship.

**You were right early — independently adopted since:**
- **MTP 5→3**: you called this on June 30. The 2026-07-03 garble fix landed the same default on main four days later (plus `draft_sample_method: probabilistic`). Credited in CREDITS.md.
- gmu 0.80→0.85: main's default is now 0.85.
- Wiring `MTP_NUM_TOKENS` into the speculative config: main now interpolates `${MTP_NUM_TOKENS:-3}` properly. (Heads-up: this PR's inline version would drop `draft_sample_method:"probabilistic"`, which would regress the garble fix — one reason we didn't merge as-is.)

**Tested today, neutral on current main:**
- `MAX_NUM_BATCHED_TOKENS` 8192→16384: A/B on identical workloads — c1 38.4 vs 38.5, c2 59.5 vs 58.8, c4 85.2 vs 86.0 tok/s; acceptance identical (~2.3-2.9). Your original observation was real, but it was a symptom of the MTP=5 garbling; with MTP=3 + probabilistic the batch size no longer moves spec quality. Keeping 8192.

- MAX_NUM_SEQS 6→4: superseded — main moved to 12 with the PR #5 capture-size formula.

Closing as partially-adopted/superseded. Appreciate you testing at 1M and writing up the reasoning — the MTP=3 call in particular was ahead of everyone. 🙏

Issue #2, opened 2026-07-02
CUDA device-side assert in DSpark draft-loop CUDA graph replay at long context (~58-91K tokens) kills engine; reproducible in ~45 min


paulbrav:
## Summary

Under sustained agent-style traffic, the engine dies with `CUDA error: device-side assert triggered` once a conversation's context grows large (~58K–91K tokens observed). With `CUDA_LAUNCH_BLOCKING=1`, the faulting launch is the **DSpark proposer's captured draft-loop CUDA graph replay** (`dspark_proposer.py::_run_draft_for_current_context` → `torch.cuda.graphs.replay`). The crash is reproducible in ~40–50 minutes with a synthetic agent-loop workload.

This defeats the recipe's core value for us: the 1M-context profile crashes two orders of magnitude below its advertised window, so long-context agent use (the reason to run this recipe at all) is not viable. Compacting agent context to stay under the crash threshold is not an acceptable workaround.

## Environment

- 2× DGX Spark (spark-01 head, spark-02 worker), TP=2 over QSFP, NCCL socket transport
- Recipe checkout at commit `89bb82b` ("Document OpenClaw bounded bootstrap fix") — noting current main is `ac3bf77`; happy to retest on the C12 profile if you believe it addresses this
- Runtime image: `vllm-dspark-runtime:dspark-nvfp4-stage-c` built by `build-dspark-vllm-runtime.sh` from base `ghcr.io/bjk110/vllm-spark:unholy-fusion-prod-ready`
- vLLM in container: `0.21.1rc1.dev339+g1967a5627bc3`
- Validated 1M profile per the README: `max_model_len=1048576`, `MAX_NUM_SEQS=6`, `MAX_NUM_BATCHED_TOKENS=8192`, `MTP_NUM_TOKENS=5`, `kv_cache_dtype=nvfp4_ds_mla`, async scheduling enabled (engine logs "Asynchronous scheduling is enabled."), CUDA graphs FULL_AND_PIECEWISE

## Original symptom

First hit during real OpenCode agent traffic: engine death (`EngineDeadError`) on a single running request at ~58K computed tokens, with the CUDA error surfacing at async sync points (`update_async_output_token_ids`, `_get_valid_sampled_token_count`) — i.e. the async-reported stack pointed at innocent bystanders.

## Deterministic-ish reproduction (synthetic workload)

[...]

PR #4, opened 2026-07-05
Fix long-context engine crashes: gate sparse-indexer gather + clamp stale DSpark draft-KV slot ids

paulbrav:
Fixes the long-context engine deaths in #2 (`CUDA error: device-side assert triggered` / illegal memory access at ~58–142K token context, also independently reproduced by @DaveCharland on a second rig with a different recipe lineage).

## What this fixes

**Diff-size note for review:** 615 of the 683 added lines are `cache_utils.py` arriving as a full-file overlay copy (this repo's patch mechanism — the file lives in the base image, so it must be copied whole to change it). The net new logic is ~75 lines, of which ~12 are the actual kernel fix; the exact per-file hunks are in the linked issue comment for direct review.

Two stacked out-of-bounds-index bugs, full analysis with every hunk inline in [this issue comment](https://github.com/tonyd2wild/DeepSeek-v4-Flash-DSpark-1M-NVFP4-KV-2x-DGX-Spark/issues/2#issuecomment-4887968751) (kept out of the tree deliberately):

1. **`cache_utils.py` sparse-indexer consumer** (new overlay file): the Triton kernel gathered `block_table` for padding tokens carrying `torch.empty` garbage from the persistent `topk_indices_buffer`. Fix: gate the gather on `is_valid_token` and bound block numbers by a new optional `num_blocks` parameter, passed from both `sm120.py` call sites (`flashmla.py`'s caller is untouched — the parameter defaults to unbounded).
2. **`dspark.py` stale draft-KV slot ids** — the load-bearing one: the draft sliding-window KV ring buffer (`main_kv_cache`, one row per `MAX_NUM_SEQS` slot) is gathered by per-request `slot_index` ids that can go stale across request condensation. A stale id ≥ num rows trips the `indexSelectSmallIndex` assert and kills the worker. Fix: `_guard_slot_index()` clamps on-device at all three gather sites, logging the offending value on eager paths.

The clamp is draft-path-only (a clamped slot yields a bad draft token the target model rejects at verification — one wasted speculation, never output corruption; measured acceptance unchanged) and CUDA-graph-safe (pure device op in the captured graph; host-side logging gated on `is_current_stream_capturing()`).

A `DSPARK_SLOT_CLAMP` compose env (default `1`) is included as a kill switch so anyone can reproduce the A/B on their own rig without rebuilding.

## Validation (synthetic workload only)

2× DGX Spark TP=2, 1M profile, `thinking:true`, seed-pinned agent-loop harness (6 conversations growing past 100K ctx with churn/aborts):

| config | outcome |
|---|---|
| baseline ×7 runs | engine death in 76–142 min |
| both fixes (run 8) | **full 4 h clean**, 117.5K ctx, acceptance 37.2% cum. (per-pos 72/49/32/20/13%) |
| both fixes (run 9) | **full 4 h clean**, 115.8K ctx |
| clamp reverted (`DSPARK_SLOT_CLAMP=0`) | **identical assert at ~120 min**, at a 6→5 batch-condensation tick |

Plus ~8 h clean under the protected config since. Remove the clamp → crash returns; restore → clean, i.e. causal proof, not correlation.

Not a duplicate: no open PR addresses this (only #1, a perf tune). The remaining producer-side root cause (what hands out the stale slot id during condensation) is documented in the doc for follow-up.

## Disclosure

AI-assisted investigation and patches (Claude); human-reviewed end to end, all validation run by the human operator. All data synthetic — no real conversation traffic.

### DIFF
--- a/docker-compose.dspark.yml
+++ b/docker-compose.dspark.yml
@@ -23,6 +23,9 @@ services:
       TRANSFORMERS_OFFLINE: "${TRANSFORMERS_OFFLINE:-0}"
       HF_HUB_DISABLE_XET: "${HF_HUB_DISABLE_XET:-1}"
       VLLM_CACHE_ROOT: /cache/huggingface/vllm-cache
+      # Long-context crash fix kill-switch (see docs/LONG_CONTEXT_CRASH_FIX.md).
+      # 1 (default) = clamp stale DSpark draft-KV slot ids; 0 = detect+log only.
+      DSPARK_SLOT_CLAMP: "${DSPARK_SLOT_CLAMP:-1}"
       VLLM_HOST_IP: "${VLLM_HOST_IP:-}"
 
       VLLM_ALLOW_LONG_MAX_MODEL_LEN: "${VLLM_ALLOW_LONG_MAX_MODEL_LEN:-1}"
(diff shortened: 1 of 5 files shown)


tonyd2wild:
Going to look on this now thanks brother

PR #7, opened 2026-07-09
Ship proposer fixes via the image only; auto-rebuild when the overlay is stale (fixes cross-version propose() crash)

robotnursenyc:
## Problem

The compose file bind-mounts `recipe/vllm/v1/spec_decode/dspark_proposer.py` over the image's own copy (the concurrency-fix delivery from 90ab8e2). That mounted copy is version-matched to one specific vLLM build — so the moment the `dspark-nvfp4-stage-c` tag is rebuilt against a newer vLLM (whose model runner passes `req_ids=` to the proposer), **every speculation step throws** `TypeError: DSparkProposer.propose() got an unexpected keyword argument 'req_ids'`, the EngineCore dies, and all completions 500 until manual intervention. This is exactly the cross-version hazard 748f034's README note warns about — this PR makes it structurally impossible instead of documented.

This took down a production 2×DGX-Spark rig on 2026-07-08 within hours of an image rebuild. A second rig was one `git pull` away from the same outage (its checkout predated the mount; current `main` would have introduced it under its rebuilt image).

## Change

- **Drop the proposer bind-mount from compose and delete the stale `recipe/vllm/` copy.** `recipe/overlay/` (baked into the image at build) is the single source of truth; a second copy exists only to drift.
- **`start-deepseek-v4-flash-dspark.sh` now verifies the image against the overlay**: it sha256-compares every file in the Dockerfile's COPY list against the image and runs `build-dspark-vllm-runtime.sh` when they diverge (`SKIP_OVERLAY_CHECK=1` opts out). Rigs whose image predates the current overlay therefore cannot lose the concurrency fix when the mount disappears — the check lives in the start script (not update-and-restart.sh) so it takes effect on the *first* post-pull run, since update-and-restart execs the start script fresh from disk.
- Stop scp-ing the deleted copy to the worker; update the README version note, the guard-applier docstring, and the Kai update note.

`scripts/apply-nonuniform-guard.py` stays, for people running foreign prebuilt images.

## Validation (two independent 2×Spark rigs)

- Rig 1 (the crashed one): mount removed → boots clean, concurrency 1/2/4/6 passes with zero bad outputs, DSpark acceptance healthy.
- Rig 2: the staleness check fired on its first start (its image was missing 4 overlay files, including the Patch 3 scheduler fix and the 1M-context int64 fix), auto-rebuilt in ~2 min via layer caching, came back serving with the overlay verified byte-identical in the image. Concurrency clean before and after.

## Related observation (follow-up material, not in this PR)

`.env.dspark.example` ships `MAX_NUM_SEQS=6` + `MTP_NUM_TOKENS=5`, which requests `6*(5+1)=36` cudagraph rows and triggers `WARNING Truncating max_cudagraph_capture_size to 32` — full spec batches then run off-graph (the ca11f3f "seqs6 MUST use spec 3" rule, which DEFAULT-CONFIG.md later overrode). Keeping `MAX_NUM_SEQS*(MTP_NUM_TOKENS+1) <= 32` and dropping to MTP=3 measured **+22% single-stream and +21-39% at C6** (prose/agent workload, temp 0, warm) on both rigs, with code single-stream flat. Happy to send that as a separate PR if wanted.


tonyd2wild:
Merged to main in 0081289 (merge had a README conflict with the newer intro/TL;DR section, resolved keeping your version-note rewrite). Thank you @robotnursenyc — this is the right structural fix: dropping the bind-mount + the second proposer copy kills the whole cross-version drift class, and the overlay-staleness rebuild guard closes the 'image predates overlay' gap (including the Patch 3 scheduler fix). Verified on our 2×Spark rig: baked proposer carries req_ids, no proposer mount, concurrency clean. The two-rig validation writeup made this an easy merge — appreciated.

PR #14, opened 2026-08-01
Add sparkrun recipe for one-command reproducible deployment (closes #9)

### LINKED ISSUE #9: [Feature Request] Add support for Sparkrun and Spark-Arena
Hello @tonyd2wild 

please add support and recipes for [sparkrun](https://github.com/spark-arena/sparkrun).

Why, whats the benefit?
That way anybody can run your recipe with your settings and replicate your success.
We can then use [llama-benchy](https://github.com/eugr/llama-benchy) and publish reports and results on [Spark-Arena](https://spark-arena.com/leaderboard).

Sources:
- sparkrun 
  - Tool for easy recipe-based provisioning
  - https://github.com/spark-arena/sparkrun
- Spark-Arena
  - LLM Performance Leaderboard - Compare Large Language Model performance on NVIDIA DGX Spark infrastructure. Real benchmarks, - real hardware, real results. 
  - https://spark-arena.com/
- llama-benchy 
  - Benchmarking Tool
  - https://github.com/eugr/llama-benchy

Regards,
jammsen

jammsen:
Closes #9

Follow-up to the recipe I posted in the issue thread — this is the same thing, brought up to date for the 0731 release and re-validated end to end on my 2x DGX Spark pair yesterday. Self-contained `sparkrun/` dir (recipe, minimal README, warm-up + speed-test script), your main README untouched — link it from wherever you like, or not at all.

What changed vs the version in the thread:

- **k=5**, per your fix. You were right about the +24%, and then some (numbers below).
- **Patch 4 comes in automatically** — the recipe pins the repo tarball at `d728fae` and copies the overlay, so the shared-expert loader fix is just there. The pre_exec also fail-fast greps for Patch 3 and Patch 4 after applying, so if the pin and the base image ever drift apart it refuses to serve instead of running the half-speed stock loader silently.
- **C6 / gmu 0.78 profile** from your DEFAULT-CONFIG "CURRENT BEST", and `--max-cudagraph-capture-size` left unset so vLLM derives it (36 = 6x6, confirmed in the boot log).
- **`apply-nonuniform-guard.py` is no longer run.** Turns out stacking it on top of the overlay's own proposer (which handles ragged batches natively) was exactly what produced the `Drafted: N, Accepted: 0` windows I mentioned back in July. Self-inflicted on my side — the guard is for foreign images, your docs even say so.

Measured 2026-08-01, warm, single-stream unless noted, tokens counted via `usage.completion_tokens`:

| workload | tok/s |
|---|---:|
| Protocol Starfall prompt (the video demo, 3603 tok) | **71.7** (was 54.7 on my k=3 slow path; video showed 50.9) |
| bulk SQL INSERTs, 1500 tok (~100% draft acceptance) | **82–83** |
| checker.ts corpus (llama-benchy), c1 | 46.3 |
| checker.ts, aggregate c6 | 74.8 |
| checker.ts, c1 at 32K depth | 45.1 |

KV pool came up at 1,671,191 tokens (1.59x at 1M) on this boot. Warm-up matters exactly like you documented, so the bundled `speedtest-starfall.sh` warms before it measures, and the README repeats your stream-chunks-vs-tokens warning because I nearly fell into that one too.

Happy to adjust naming/layout however you want it.

### DIFF
--- a/sparkrun/README.md
+++ b/sparkrun/README.md
@@ -1,0 +1,86 @@
+# sparkrun recipe — DeepSeek V4 Flash 0731, 1M NVFP4 KV, 2x DGX Spark
+
+One-command, copy-paste reproducible deployment of this repo's setup via
+[sparkrun](https://github.com/spark-arena/sparkrun). No image build: the
+recipe pulls the public base image (digest-pinned, so it can never silently
+change) and rebuilds the `dspark-nvfp4-stage-c` runtime inside every
+container at start — overlay, stage A/B/C patches, Patch 3 and Patch 4
+included, pinned to commit `d728fae` of this repo. It fail-fast-verifies
+Patch 3 and Patch 4 actually landed before serving, so you can't
+accidentally run the half-speed stock loader.
+
+Everything here was deployed and measured on a real 2x DGX Spark pair on
+2026-08-01. For tuning, patch history and troubleshooting, the
+[main README](../README.md) is the reference — this file is just how to run
+it.
+
+## Quick start
+
+Once, on your head node (~200 GB free disk per node):
+
+```bash
+uvx sparkrun setup     # wizard: cluster, SSH mesh, ConnectX-7 fabric detection
+```
+
+Then:
+
+```bash
+sparkrun run ./deepseek-v4-flash-0731-dspark-nvfp4-1m-vllm.yaml
+```
+
+sparkrun syncs the image (~13 GB) and model (167 GB, first time only) to
+both nodes, injects per-host NCCL/RoCE settings, launches the worker
+headless and the head with the API on port 8888. First-ever boot is ~9 min,
+warm boots ~5 min. `Ctrl+C` detaches; `sparkrun logs` / `status` / `stop`
+manage it afterwards.
+
+Verify the boot:
+
+```bash
+curl -fsS http://<head-ip>:8888/v1/models   # expect "max_model_len": 1048576
+sparkrun logs deepseek-v4-flash-0731-dspark-nvfp4-1m-vllm | grep -E "B12X|KV cache size"
+# want:  Using 'B12X' Mxfp4 MoE backend     (missing = half-speed fallback)
+#        GPU KV cache size: ~1.5-1.7M tokens (varies per boot, that's normal)
+```
+
+## Measured (2x DGX Spark, TP=2, warm, 2026-08-01)
+
+| workload | tok/s |
+|---|---:|
+| Protocol Starfall prompt (the video demo; video showed 50.9 live) | **71.7** |
+| bulk SQL INSERTs (~100% draft acceptance) | **82–83** |
+| hard code corpus (checker.ts), single-stream | 46.3 |
+| hard code corpus, aggregate at c6 | 74.8 |
+| hard code corpus, single-stream at 32K depth | 45.1 |
+
+Decode is acceptance-bound and content-driven — the spread above is one
+server on one config, exactly as the main README describes.
+
+## Speed test
+
+```bash
+./speedtest-starfall.sh http://<head-ip>:8888
+```
+
+Warms the engine (first requests after boot or ~30 min idle run ~30% slow),
+then measures the Starfall prompt the right way: `stream: false` +
+`usage.completion_tokens`. Don't count SSE chunks — under spec decode vLLM
+emits one chunk per decode *step*, so stream-delta counting reports steps/s
+and under-reads by the acceptance length.
+
+## Benchmark
+
+The recipe ships a coding-flavored `benchmark:` block (llama-benchy's
+default corpus is a novel, which under-reports this model for code work):
+
+```bash
+sparkrun benchmark ./deepseek-v4-flash-0731-dspark-nvfp4-1m-vllm.yaml --skip-run
+```
+
+runs depth 0/32K x concurrency 1/2/6 (~11 min) against a running server.
+Corpus A is TypeScript compiler internals (conservative); switch to
+boilerplate-heavy HTML for the upper bound with
+`-b book_url=https://raw.githubusercontent.com/whatwg/html/88ae68cb961651f0f92c5d2046049f53ecdfc6cf/source`.
+Two caveats: warm up first, and ignore the depth-32K cells at concurrency >1
+unless you raise `-b tg=1024` — at the default `tg=128` those cells measure
+the chunked-prefill storm, not decode.
--- a/sparkrun/deepseek-v4-flash-0731-dspark-nvfp4-1m-vllm.yaml
+++ b/sparkrun/deepseek-v4-flash-0731-dspark-nvfp4-1m-vllm.yaml
@@ -1,0 +1,320 @@
+# DeepSeek V4 Flash 0731 (official) DSpark — 1M context, NVFP4 KV — 2x DGX Spark
+#
+# sparkrun port of:
+#   https://github.com/tonyd2wild/DeepSeek-v4-Flash-0731-DSpark-1M-NVFP4-KV-2x-DGX-Spark
+# (pinned to commit d728faee9f5a8d5ebafe7bc44bca6c5d8d0d192f, 2026-07-31)
+#
+# This is the recipe for DeepSeek's OFFICIAL `DeepSeek-V4-Flash-0731` release,
+# contributed via issue #9. See sparkrun/README.md for quick start and
+# measured results; the main repo README stays the reference for everything
+# else (tuning, troubleshooting, patch history).
+#
+# Fully reproducible, no manual image build: the recipe pulls the public
+# DSpark-enabled base image and reproduces tonyd2wild's
+# `vllm-dspark-runtime:dspark-nvfp4-stage-c` image at container start via
+# pre_exec (base overlay -> nvfp4 stage A/B/C patches). The overlay at this
+# pin already contains ALL known fixes:
+#   * Patch 3 (scheduler cold-prefill garble root fix; without it 44/44
+#     cold ~20K-token prompts garbled, warm requests never fail — invisible
+#     until production)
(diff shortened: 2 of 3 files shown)


AndreasKunar:
@jammsen  - Thanks a lot, Sebastian (und ein Servus aus Wien)! You made my day, after trying unsuccessfully to patch it myself to 0731 (my experiments ran much slower).

It works perfectly for me. My quick & dirty llama-benchy results are: pp2048 is ~1700 ± 60 t/s, tg32 ~36 ± 3 t/s  

FYI - I'm running my Sparks normally limited to 2GHz clock, to prevent overheating and significantly reducing max. energy-consumption (see https://forums.developer.nvidia.com/t/cooler-gb10-temps-almost-no-performance-lost/372662). This reduces pp2048 here to \~1550 ± 21 t/s (\~ -10% on pp, while saving \~25% on power and running cooler / with sustainable load-performance). tg is not impacted, because it's memory-bandwidth limited.

jammsen:
> @jammsen - Thanks a lot, Sebastian (und ein Servus aus Wien)! You made my day, after trying unsuccessfully to patch it myself to 0731 (my experiments ran much slower).
> 
> It works perfectly for me. My quick & dirty llama-benchy results are: pp2048 is ~1700 ± 60 t/s, tg32 ~36 ± 3 t/s
> 
> FYI - I'm running my Sparks normally limited to 2GHz clock, to prevent overheating and significantly reducing max. energy-consumption (see https://forums.developer.nvidia.com/t/cooler-gb10-temps-almost-no-performance-lost/372662). This reduces pp2048 here to ~1550 ± 21 t/s (~ -10% on pp, while saving ~25% on power and running cooler / with sustainable load-performance). tg is not impacted, because it's memory-bandwidth limited.

TG is mostly Memory-Bandwidth limited, aside from PCIe-Bus-Speeds that affect this also.
Where big omph is needed it PP.

Why and how would Power-Limit your Spark?

AndreasKunar:
> ...
> 
> Why and how would Power-Limit your Spark?

@jammsen - for technical details see this [post on the NVTDIA DGX Spark forum ](https://forums.developer.nvidia.com/t/cooler-gb10-temps-almost-no-performance-lost/372662). Net:
- According to NVIDIA, its Ideal operating temperature is - 5°C to 30°C (41°F to 86°F). Even with USB-fans, my cluster got hotter at home, when it's now 30°+ outside (the room with my Spark-cluster has no AC). So sometimes, I got random, uncontrolled Spark shutdowns.
- Reducing the sm clock to 2000 with `sudo nvidia-smi -lgc 0,2000` reduces total Spark power consumption by ~ 25% and reduces the GB10 GPU's power-consumption by up to 50% and its temperature significantly (no more 85+°C GPU-temp). So, no shutdowns anymore, even with extreme load, and almost no performance impact.
- And your work on enabling this fast model-speed, enabled me to easily have the 10% pp reserve to give away. Thanks!

This was the meaning of my posting.

jammsen:
> > ...
> > Why and how would Power-Limit your Spark?
> 
> @jammsen - for technical details see this [post on the NVTDIA DGX Spark forum ](https://forums.developer.nvidia.com/t/cooler-gb10-temps-almost-no-performance-lost/372662). Net:
> 
> * According to NVIDIA, its Ideal operating temperature is - 5°C to 30°C (41°F to 86°F). Even with USB-fans, my cluster got hotter at home, when it's now 30°+ outside (the room with my Spark-cluster has no AC). So sometimes, I got random, uncontrolled Spark shutdowns.
> * Reducing the sm clock to 2000 with `sudo nvidia-smi -lgc 0,2000` reduces total Spark power consumption by ~ 25% and reduces the GB10 GPU's power-consumption by up to 50% and its temperature significantly (no more 85+°C GPU-temp). So, no shutdowns anymore, even with extreme load, and almost no performance impact.
> * And your work on enabling this fast model-speed, enabled me to easily have the 10% pp reserve to give away. Thanks!
> 
> This was the meaning of my posting.
Gotcha:

Followup-Question:
What Sparks do you have? 
Afaik the best Thermal-Solution Spark is currently MSI, then Asus, then everyone else and then dead last Nvidia. My Sparks were standing ontop of each other but that was getting fricking hot. Thats why i used my 3D-Printer to print 4 sides and 4 tops of these: 
https://makerworld.com/en/models/2449541-dgx-spark-dell-pro-max-gb10-stack?from=search#profileId-2689137

They look like this:
image
(The 3 P above each other parts are the Vertical Elements)

And after arranging them like this, i let them benchmark for over 5 hours, never got over 82-85°C and my room also has not AC, they just need breathing room, no matter what Asus or Nvidia say.

Also 
>  (und ein Servus aus Wien)!

Hi! 😎

AndreasKunar:
> ...
> 
> Followup-Question: What Sparks do you have? Afaik the best Thermal-Solution Spark is currently MSI, then Asus, then everyone else and then dead last Nvidia. My Sparks were standing ontop of each other but that was getting fricking hot. Thats why i used my 3D-Printer to print 4 sides and 4 tops of these: https://makerworld.com/en/models/2449541-dgx-spark-dell-pro-max-gb10-stack?from=search#profileId-2689137
> 
> They look like this: ...
> 
> And after arranging them like this, i let them benchmark for over 5 hours, never got over 82-85°C and my room also has not AC, they just need breathing room, no matter what Asus or Nvidia say.
> 
> Also
> 
> > (und ein Servus aus Wien)!
> 
> Hi! 😎

@jammsen - thanks, your solution looks great!
I have 2x ASUS Ascent GX10 (they were the cheapest, EUR 3000 in Feb). My 2 simple USB-fans from amazon on top are OK for moving the heat-dissipation - from the power-supplies, the SSD-enclosures, and GX10 exhausts - out+up from the rack. The GX10 has its intakes front and front-bottom, this helps. But it's still not good enough if the outside temp is >35°, unless I throttle the GB10. Before throttling, I still sometimes had the GX10 powering off.

IMG_8632

jammsen:
Are those usb-c to nvme adapters for mass-storage of models? 😎 
I think i need to do that too 🤣

AndreasKunar:
> Are those usb-c to nvme adapters for mass-storage of models? 😎 I think i need to do that too 🤣

@jammsen - important: you need USB3.2x2 (20Gbps) enclosures - not modern USB4/Thunderbolt4 or 5 (for me, these had USB-2 speed on the Spark). Best also with fans to keep the SSD from overheating. For the Spark, I can recommend GRAUGEAR External M.2 SSD Enclosures for NVMe SSD M.2 20Gbps.

jammsen:
> > Are those usb-c to nvme adapters for mass-storage of models? 😎 I think i need to do that too 🤣
> 
> @jammsen - important: you need USB3.2x2 (20Gbps) enclosures - not modern USB4/Thunderbolt4 or 5 (for me, these had USB-2 speed on the Spark). Best also with fans to keep the SSD from overheating. For the Spark, I can recommend GRAUGEAR External M.2 SSD Enclosures for NVMe SSD M.2 20Gbps.

@AndreasKunar 
Cooling for the SSDs? Oh wait, do you actively host the llms from that SSDs? 
I personally  wouldnt do that, a) its to slow and b) i almost use entirely vLLM to the model is already in RAM. 

I also would only use them only to archive/keep the models that i dont need currently. OR just to get more space for new ones to evaluate. I have the 1TB version of the sparks, by choice. Its MUCH MUCH cheaper to buy the 1TB one and if needed buy a 4TB 2242/2230 and clone the 1TB to the 4TB, extend an use it. You still safe between 500-1200€ that way per Spark.

Though i could just use my NAS via the 10GbE nic or the USB-to-M2-Version, which might be faster. 

Many options here 😎

tmolteno:
Amazing work. Needs a minor revision for second time startup. VLLM_ENGINE_READY_TIMEOUT_S: "3600" to avoid getting torch.distributed timeouts during startup (it defaults to 600 seconds.

jammsen:
> Amazing work. Needs a minor revision for second time startup. VLLM_ENGINE_READY_TIMEOUT_S: "3600" to avoid getting torch.distributed timeouts during startup (it defaults to 600 seconds.

Do a PR then 😎

tmolteno:
PR #19 done. Thanks for the nudge. Used qwen code and my local deepseek-v4-flash-0731 to do the PR :)

tonyd2wild:
Merging. It's self-contained under `sparkrun/`, doesn't touch the main README, and it serves the no-image-build path that the compose launcher deliberately doesn't — different audience, not a competing one.

The technical content is right and clearly came from actually running it: `k=5`, `max_num_seqs 6`, gmu 0.78, `max_cudagraph_capture_size` deliberately unset, and correctly *not* running `apply-nonuniform-guard.py` on top of the overlay.

The `pre_exec` gate deserves specific credit. Grepping for `shared_experts.gate_up_proj` and refusing to serve if Patch 4 is missing is exactly the right call — without that patch the 0731 checkpoint runs at roughly half decode speed with **perfect output quality**, so a benchmark run would silently publish the slow number and nobody would notice. Fail-closed beats a comment.

Two things to note in the README so nobody is surprised later:

1. The `pre_exec` reconstructs the overlay by `sed`-ing the `<<'PY'` heredocs out of `recipe/nvfp4/Dockerfile.stage-*`. That's a coupling to Dockerfile *formatting*, not an interface — I'm adding a comment in those Dockerfiles noting the heredoc markers are load-bearing for this recipe.
2. The `d728fae` pin is current `main` today, so it needs a bump each release or sparkrun users silently get an old overlay. Adding it to the release checklist.

One suggestion rather than a blocker: pin the orchestrator too — `uvx sparkrun@0.3.1` rather than bare `uvx sparkrun`. The setup wizard configures an SSH mesh across the user's nodes, which is a lot of trust for a latest-version resolve. Provenance checks out (PyPI `sparkrun` 0.3.1 matches `spark-arena/sparkrun`, same author), so this is about reproducibility, not suspicion.

Closes #9. @AndreasKunar — thanks for independently running it unedited; your `chown -R 1000:1000` fix is in the README troubleshooting section.

PR #17, opened 2026-08-03
Refactor DeepSeek V4 Tool Parser and Encoding

hhackbarth:
- Updated `start-deepseek-v4-flash-dspark.sh` and `status-deepseek-v4-flash-dspark.sh` to derive API and chat URLs after sourcing environment variables, allowing for dynamic port configuration.
- Introduced `deepseek_v4.py` and `deepseek_v4_encoding.py` for custom tokenizer and encoding logic specific to DeepSeek V4, enhancing message formatting and tool call handling.
- Added `deepseekv32_tool_parser.py` to implement a new tool parser for handling DSML function calls, including robust parameter parsing and streaming support.
- Improved error handling and logging throughout the tool parsing process to ensure better traceability and debugging.


tonyd2wild:
The bug is real. I reproduced it on my live server before reviewing the code, so let's start there.

Streaming request with one tool defined, plain text answer, 40 max tokens:

```
total data frames          : 5
frames with "tool_calls":[] : 3
```

Every content chunk carries `"tool_calls":[]` once a request has tools. Your diagnosis is exactly right, and the cause is the explicit construction in stock `extract_tool_calls_streaming`:

```python
return DeltaMessage(content=content, tool_calls=list(tool_call_deltas.values()))
```

On a plain-text delta `tool_call_deltas` is `{}`, so that passes `[]` explicitly, which marks the Pydantic field set, and with `exclude_unset=True` the key ships on the wire. Not OpenAI format, and `[]` is truthy in JS. That fully explains the VS Code symptom you described.

Thank you for finding it. That has been silently breaking agent clients against this repo and I did not know.

**I am not merging this as written, and here is the specific reason.**

The fix is about four lines. This PR is 287 deleted lines against stock. Diffing your file against stock vLLM at `00eb7ce`, which is what the runtime is built from, it deletes the entire incremental streaming engine: `_process_streaming_buffer`, `_json_escape_string_content`, `_add_tool_call_delta`, `_begin_streaming_tool_call`, `_append_param_prefix`, `_append_json_param_value`, `_param_types_for_name`, `_can_stream_raw_param`, `_should_buffer_wrapper_param`, `_finish_buffered_param`, `_close_streaming_tool_call`, `_get_param_config`, both partial regexes, and all buffer state. It replaces streaming with buffer-until-complete.

That trade is not mentioned anywhere in the PR body, which describes the change as adding "robust parameter parsing and streaming support" and "improved error handling." The diff does close to the opposite on both counts.

Four consequences I can't take on a live cluster:

1. **Stream stalls.** Nothing is emitted for a tool call until `</｜DSML｜invoke>` closes. An agent writing a 4KB file body through a tool parameter now produces zero SSE bytes for that whole span. At DSpark decode rates that is long enough for clients with idle timeouts to drop the connection.
2. **O(n²) per response.** `invoke_complete_regex.findall(current_text)` and `partial_tag_overlap` run over the entire accumulated text on every token, in the shared async output loop. Stock scans an incremental buffer.
3. **Content after a tool block is dropped.** `_extract_content` uses `current_text.index(self.tool_call_start_token)`, the first occurrence, so `sendable_idx` freezes permanently once a block opens. Stock clears `_in_tool_calls` at the close token and resumes. Off-template for V4 today, but it is a real loss.
4. **Error handling narrowed.** Stock's streaming path had `logger.exception` on finalize failure. The new streaming path has no try/except at all, so a malformed-output exception propagates into the output loop.

**What I would merge today:** the minimal guard, keeping stock's engine intact and changing only the return.

```python
if tool_call_deltas:
    return DeltaMessage(content=content, tool_calls=list(tool_call_deltas.values()))
return DeltaMessage(content=content)
```

That kills the `"tool_calls":[]` frames with zero granularity change, zero O(n²), zero content-loss edge case.

**Also worth knowing:** upstream at `00eb7ce` ships `tests/tool_parsers/test_deepseekv32_tool_parser.py` (1222 lines) and `test_deepseekv4_tool_parser.py` (352 lines), including `test_streaming_matches_non_streaming`, `test_multiple_tools_streaming`, `test_type_conversion_in_streaming`, and a chunked case asserting DSML start-marker fragments must not leak. They are pure Python, no GPU. I would expect several granularity assertions to fail against this version. Worth running either way, and that is my gap as much as yours since this repo has no tests directory at all.

**Happy to take separately, they look correct:**
- The `start-` / `status-` / `smoke-` URL derivation fix. Those were pinned to 8888 before `.env` was sourced. Clear bug, unrelated to the parser.
- The `agent_sanity_bench.py` empty-output gate. Note it now fails closed on empty completions, so with thinking on at the default `MAX_TOKENS=256` it will report failures that used to pass silently.
- The reasoning-effort restoration. Your read is right, stock only carries the `max` constant and the normalizer collapsed everything else to `high`, which injected nothing. It is inert for this deployment since we run thinking off, but it is correct.

**One conflict:** your `GLOO_SOCKET_IFNAME` / `TP_SOCKET_IFNAME` change overlaps open PR #13. Those need to be one change, not two. I have asked over there for the same thing.

**One caution:** the compose change reintroduces `--override-generation-config`. You document the footgun well, and it defaults empty. But that flag was removed after the 2026-07-03 garble incident and I would rather it stay gone unless there is a reason to bring it back.

Split the parser change down to the guard and I will merge it quickly. The find itself is genuinely valuable.

tonyd2wild:
Correction. I owe you one, and it is on the central point of my review.

I said your PR "deletes the entire incremental streaming engine" and framed that as a regression from stock. I was diffing your branch against upstream vLLM at `00eb7ce`. That was the wrong baseline, because it is not what this repo actually runs.

I pulled the parser out of the running container tonight:

```
installed deepseekv32_tool_parser.py : 307 lines
upstream 00eb7ce                     : 555+ lines
```

And its own docstring says:

> Uses a buffer-until-complete-invoke strategy: tokens are buffered
> until a complete invoke block is available, then parsed and emitted
> in one shot.

Then I checked for your distinctive methods versus stock's:

```
_extract_content            deployed=2  your PR=2
_extract_delta_tool_calls   deployed=2  your PR=2
_reset_streaming_state      deployed=2  your PR=2
invoke_complete_regex       deployed=3  your PR=3

_process_streaming_buffer   deployed=0   (stock's engine)
_append_json_param_value    deployed=0   (stock's engine)
_begin_streaming_tool_call  deployed=0   (stock's engine)
```

**The base image already ships buffer-until-complete. Stock's incremental engine is not in it and never was.** So the three things I listed as regressions you would introduce, stream stalls on long arguments, content dropped after a tool block, and loss of incremental argument emission, are the existing behaviour on this deployment. You were not proposing them. You were vendoring what is already running and adding the guard.

I should have read the container before reading the diff. Apologies for the noise.

**What I verified about the guard itself**, in a CPU-only container off the running image, seed-matched, no GPU:

```
deployed parser, unmodified : 55 passed, 2 failed
deployed parser + the guard : 55 passed, 2 failed
```

Same two failures both times, `test_streaming_emits_incremental_argument_chunks` and `test_emits_arguments_before_invoke_completes`. Both are incremental-argument tests that a buffer-until-complete parser cannot satisfy by design, so they fail identically with and without the change. **Zero regression from the guard.**

**What still stands from my review**, and I do not think any of it is controversial:

- The `GLOO_SOCKET_IFNAME` / `TP_SOCKET_IFNAME` change overlaps open PR #13. Those need to be one change.
- `--override-generation-config` was removed after the 2026-07-03 garble incident. It defaults empty here and you document the footgun well, but I would rather it stay gone without a specific reason.
- The `start-` / `status-` / `smoke-` URL derivation fix is a clear bug fix and I will take it.
- The reasoning-effort restoration is correct.

**What changes:** I am no longer asking you to cut this down to four lines. Please split it so the socket ifname change goes with #13, and I will merge the parser work. If you would rather I just take the guard first so the wire format is fixed today and the rest lands after, say so and I will do that instead.

The bug you found is real and it is live: I measured 3 of 5 streamed frames carrying `"tool_calls":[]` on this cluster. Thank you for it, and sorry for the mischaracterisation.

hhackbarth:
Both of your suggestions addressed in 134c695, docker-compose.dspark.yml only, +4 −23.

Socket ifnames removed. They belong with #13, which covers the same fix plus .env.dspark.example and a docs/PORTABILITY.md. I'll leave a note over there with what we measured independently, so the reasoning isn't lost when this drops out here.

--override-generation-config removed entirely — flag, env var, and the argument-array plumbing. You're right that nothing in this PR needs it. We were using it locally for a server-side temperature: 0.5; we've given that up too, so our deployment now matches upstream. Clients that want a non-default temperature send it per request, which took precedence anyway.

I kept DEFAULT_CHAT_TEMPLATE_KWARGS. It defaults to the previous hardcoded '{"thinking":false}', so no behaviour change, and without it the restored three-level reasoning_effort is only reachable per request rather than from configuration. Happy to drop it as well if you'd rather the parameterisation land separately.

Untouched: the tool-parser guard, the tokenizer reasoning_effort restoration, the start-/status-/smoke- URL derivation, and the agent_sanity_bench empty-output gate.

On the tests — you're right that they're the gap. Your seed-matched run (55 passed / 2 failed, identical with and without the guard, both failures inherent to buffer-until-complete) is better evidence than anything I produced; I verified end-to-end on the live server instead: no empty-tool_calls deltas, real tool calls unchanged with full arguments, complete agent loop, and VS Code Insiders working without the proxy I'd been using as a workaround. If you want the upstream tests/tool_parsers/ files vendored in as a follow-up PR, I'm glad to do that — it's CPU-only and would give this repo a regression net.

And thanks for going back and re-checking against the container. That correction saved me from arguing a point I'd have lost on the wrong baseline too.

tonyd2wild:
Thanks — the analysis behind this is sound and I verified the load-bearing claims independently rather than taking the description on faith.

**The blocking question is resolved.** I checked the in-image parser on our Spark:

```
deepseekv32_tool_parser.py    307 lines, 0 occurrences of _process_streaming_buffer
  → this fork predates upstream #42879, so replacing the file reverts nothing

deepseekv4_tool_parser.py:14  class DeepSeekV4ToolParser(DeepSeekV32ToolParser)
                        :22       tool_call_start_token = "<｜DSML｜tool_calls>"
                        :23       tool_call_end_token   = "</｜DSML｜tool_calls>"
  → the V4 subclass overrides the block tokens, so patching the v3.2 base does reach V4
```

That second one matters: on a first read the `function_calls` token in the v3.2 base looks like it contradicts the encoder's `tool_calls`, and I nearly rejected this PR over it. The subclass override makes it correct.

**Confirmed good:**
- The three-level `REASONING_EFFORT_PROMPTS` is byte-identical to DeepSeek's own `encoding/encoding_dsv4.py` for the 0731 checkpoint (`low`/`high`/`max` = 0/476/526 chars). So `reasoning_effort=max` really is giving us DeepSeek's *high* text today, and true max is unreachable. Good catch.
- The `tool_calls: []` omission on content-only deltas is a real fix and is still **not** fixed upstream.
- The port-derivation fix in `start-`/`status-`/`smoke-*` is a genuine bug — today a non-default `VLLM_PORT` starts fine and then times out waiting on 8888.
- Gloo/TP `SOCKET_IFNAME` pinning is correct (and overlaps #13, now merged).

**Measured** on a 23-shape DSML differential harness, current fork-era parser vs this PR, both through the real `DeepSeekV4ToolParser`:

| | current | this PR |
|---|---|---|
| bogus `tool_calls: []` deltas | 53 | **0** |
| non-streaming arg typing | `{"days":"3","metric":"true","coords":"[1, 2]"}` | `{"days":3,"metric":true,"coords":[1,2]}` |
| parse rate (non-streaming / streaming) | 18/20, 19/20 | 18/20, 19/20 |

**Changes I'd like before merging:**

1. **Split this PR.** The compose/socket/port/bench changes are unrelated to the parser. Three are trivially reviewable; one is a 1,185-line vendored-file drop. Splitting lands the easy value immediately.
2. **Fix the Dockerfile comment.** It says the stock build "mapped EVERYTHING except max/xhigh to high" — that's upstream `main`, not this fork. As written it'll mislead the next reader.
3. **`py_compile` isn't enough.** It validates syntax, not imports; a bad import gives a green build and a server that won't boot. Add an import smoke check to the overlay stage.
4. **Release-note two behaviour changes:** `reasoning_effort="max"` will inject a materially stronger prompt than today, and `string="true"` params are no longer schema-coerced in streaming (`{"days": 3}` → `{"days": "3"}` when the model mislabels a numeric). The latter is a strict per-shape regression, though it does make streaming and non-streaming consistent.

**One thing worth stating plainly in the description:** this does **not** address the ASCII-pipe degradation (`<||DSML||tool_calls>`), which is our actual recurring field failure — every regex here is hardcoded to U+FF5C, and degraded markup still surfaces as plain-text content before *and* after. Same for markup truncated at `max_tokens`. Those are the higher-value fixes and neither is in scope here; not a criticism of this PR, just so nobody assumes it's covered.

Happy to merge the split-out parser piece as soon as 2–4 are addressed.

hhackbarth:
Split done, and 2–4 addressed. This PR is now the parser only: 2 files, +375, no tokenizer or script changes. The rest is in a separate PR #24 (`reasoning-effort-and-script-fixes`) carrying the reasoning_effort restoration, the compose parameterisation, the port derivation and the bench gate.

**Import smoke check added.** You're right that `py_compile` was insufficient. The overlay stage now imports both parsers and asserts `DeepSeekV4ToolParser` still subclasses `DeepSeekV32ToolParser` - the exact property that makes patching the base reach V4, and the one you nearly rejected this over. Build output:
```
ok vllm.tool_parsers.deepseekv32_tool_parser.DeepSeekV32ToolParser
ok vllm.tool_parsers.deepseekv4_tool_parser.DeepSeekV4ToolParser
ok DeepSeekV4ToolParser subclasses DeepSeekV32ToolParser
```
**On the Dockerfile comment** - I checked before editing, and I think the baseline in your note is off. Pulled from this fork's base image, ghcr.io/bjk110/vllm-spark:unholy-fusion-prod-ready:

```
# deepseek_v4.py:43-52
elif reasoning_effort in ("max", "xhigh"):
    reasoning_effort = "max"
else:
    reasoning_effort = "high"
```
and `deepseek_v4_encoding.py` there has only `REASONING_EFFORT_MAX`, no three-level table. So the comment did describe this fork. That said, your underlying point stands: the wording glossed over the `none` and non-string branches. I've rewritten it to name the image it was verified against and quote all four branches. That change lives in the other PR now, since that's where the tokenizer went.

**On the release notes** - the `reasoning_effort="max"` one is right and is now stated in both the commit message and the Dockerfile comment: after the patch it injects DeepSeek's max text (526 chars) instead of its high text (476), so existing callers of `"max"` get a materially stronger instruction. `"low"` and omitted render byte-identically, verified via `/tokenize`.

The streaming type-coercion one I can't reproduce as ours. Diffing the vendored file against the base image:
```
base image  307 lines
this PR     321 lines
→ 1 changed line, 2 added, 12 comment lines. Nothing else.
```
The changed line is only which branch constructs the `DeltaMessage;` no parsing, typing or schema code is touched. Your own table also shows the better typing in the this PR column, which makes me think the `current` side of that row wasn't the fork-era base image. Happy to release-note it if you can reproduce it against `unholy-fusion-prod-ready` - just want to avoid documenting a regression the diff can't produce.

**Noted on scope**: ASCII-pipe degradation and `max_tokens`-truncated markup are untouched here. Agreed those are the higher-value fixes; every regex in this file is hardcoded to U+FF5C. Not something I'd want to bolt onto this PR.

One ordering note: both PRs touch the `py_compile` list in the same Dockerfile block, so whichever lands second needs a trivial rebase. Merge this one first and I'll rebase the other.

tonyd2wild:
All four items are addressed and I verified the load-bearing ones rather than taking the description on faith.

Pulled `9ab54fc` and read the file: the change is the guard shape I asked for — `if delta_tool_calls: DeltaMessage(content=…, tool_calls=…)`, else `if content: DeltaMessage(content=content)`. 321 lines, compiles clean, `_process_streaming_buffer` absent (matches the base image, so no revert of upstream #42879). The comment block explaining *why* omitting the field matters — Pydantic `default_factory=list` + `exclude_unset=True`, `[]` being truthy in JS, the VS Code symptom — is exactly what the next reader needs.

The import smoke check is the right answer to my `py_compile` objection, and asserting `DeepSeekV4ToolParser` still subclasses `DeepSeekV32ToolParser` pins the precise property I nearly rejected this PR over.

**On the streaming type-coercion release note — you're right and I was wrong.** Your diff can't produce that behaviour change (1 changed line, 2 added, 12 comment), so my `current` column can't have been what I claimed. Withdrawn; don't release-note it.

Merging. Thank you for the find — 3 of 5 streamed frames carrying `"tool_calls":[]` had been silently breaking agent clients against this repo and I didn't know.

Restating one thing so nobody assumes otherwise: this does **not** cover ASCII-pipe degradation (`<||DSML||tool_calls>`) or markup truncated at `max_tokens`. Every regex here is hardcoded to U+FF5C. Those remain the higher-value work and neither is in scope here.

PR #19, opened 2026-08-05
sparkrun: raise VLLM_ENGINE_READY_TIMEOUT_S to 3600 for second-startup stability

tmolteno:
Follow-up to #14 (per the comment thread)

Setup: Dual DGX Spark.

Description: First run works fine, and the model is served. Second run shows this error in the container logs and torch.distributed tears the world down mid-boot, and the model serve fails.

Solution: Sets `VLLM_ENGINE_READY_TIMEOUT_S: "3600"` in `sparkrun/deepseek-v4-flash-0731-dspark-nvfp4-1m-vllm.yaml` so warm restarts of the 1M recipe get a full hour to become ready.

After some detective work, this fixed the problem.

### DIFF
--- a/sparkrun/deepseek-v4-flash-0731-dspark-nvfp4-1m-vllm.yaml
+++ b/sparkrun/deepseek-v4-flash-0731-dspark-nvfp4-1m-vllm.yaml
@@ -165,6 +165,9 @@ env:
   # and the failure log is empty.
   PYTHONUNBUFFERED: "1"
   # --- core runtime (mirrors upstream docker-compose.dspark.yml) ---
+  # Second-startup stability: engine-ready timeout defaults to 600 s, which
+  # trips torch.distributed teardown during warm restarts of this model.
+  VLLM_ENGINE_READY_TIMEOUT_S: "3600"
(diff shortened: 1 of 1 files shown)


tonyd2wild:
Merging — verified the knob is real in our fork rather than taking it on faith: `envs.py:28` `VLLM_ENGINE_READY_TIMEOUT_S: int = 600`, consumed at `v1/engine/core_client.py:614` and `:1635`, with an error message that literally tells you to raise it. So the failure mode you hit — the engine-core handshake timing out during a warm two-node start of a 156 GB model and taking `torch.distributed` down with it — is exactly what this variable governs.

Risk is minimal: it only lengthens how long the client waits for the handshake. It can't mask a crash, since a dead engine raises its own error; it only delays the "not ready" verdict. The cost is that a genuinely hung boot now hangs an hour instead of ten minutes.

Two suggestions rather than blockers:

1. **1800 would cover any 1M cold start we've observed** with half the hang time. Your call — if you've seen a real start exceed that, keep 3600.
2. The same variable should be mirrored into `docker-compose.dspark.yml` so the compose and sparkrun paths don't diverge. I'll handle that side.

PR #21, opened 2026-08-09
Patch 5: don't evaluate client stop strings inside the reasoning segment

Capicua25x:
**Patch 5: don't evaluate client stop strings inside the reasoning segment**

With think-in-prompt templates generation begins inside `<think>`, and CoT restates
phrases like `Question:`. Any harness sending stop sequences — lm-eval sends
`stop[:4]` on every request — decapitates the reasoning: `</think>` never arrives
and the parser returns `content: null`. It looks like a model failure and is a
serving one.

Per-request guard, no config. If the last prompt token is `<think>`, stops stay
dormant until `</think>`. EOS and `max_tokens` unaffected, non-thinking requests
untouched. Opt-out via `VLLM_SUPPRESS_STOPS_IN_REASONING=0`.

GSM8K n=50, temp 0.6: **8–15 nulls / 0.66–0.84 → 1 null / 0.98**. The residual
null is mechanism (B) of #18 and out of scope here.

Dogfooded four days under real agent traffic before submitting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,22 @@ Measured on 2× DGX Spark (GB10), TP=2, k=5, `unsloth/DeepSeek-V4-Flash-0731`:
 The single residual null is mechanism (B) in [#18](../../issues/18) — a marginal-stability
 reasoning runaway, unrelated to stop strings. Patch 5 does not address it.
 
+## Patch 5 and issue #18 (B): the runaway becomes *more* visible, not less
+
+Worth stating so nobody reads it as a regression. Mechanism (B) is a reasoning runaway in
+which `</think>` never arrives. Because Patch 5 keeps stops dormant until the end marker
+appears, a request in that state now has stops dormant for its whole life and runs to
+`max_tokens` — where previously a client stop string could cut it short by accident.
+
+That is correct by design: a stop string was never meant to bound reasoning, and a run
+truncated by one was returning `content: null` anyway. But the practical effect is that
+(B) shows up as a full-budget request after this patch instead of a short one, so a fleet
+that applies Patch 5 may see *reported* token usage on those requests rise. The failure
+rate does not change; only how long each failure takes to admit it.
+
+If you are measuring (B), note that stop strings are no longer a confound in either
+direction, which is the point.
+
 Apply **[Patch 4](patches/0004-dspark-shared-expert-gate-up-proj.patch)** on top of your existing
 setup. Two lines, no rebuild — bind-mount it read-only:
 
--- a/README.md
+++ b/README.md
@@ -33,6 +33,44 @@ silently drops twelve tensors, and 0731 is far more sensitive to the loss than t
 
 ### What to change
 
+## Patch 5 — stop strings must not fire inside the reasoning segment
+
+If you serve this model with a harness that sends `stop` sequences (lm-evaluation-harness
+sends `stop[:4]` on **every** request), you are almost certainly losing answers silently.
+
+vLLM's v1 detokenizer matches client stop strings against the whole output stream. With
+think-in-prompt templates, generation starts *inside* `<think>`, and chain-of-thought
+naturally restates phrases like `Question:`. The stop fires mid-reasoning, `</think>`
+never arrives, and the reasoning parser returns `content: null`. The request looks like a
+model failure; it is a serving-layer one. Hosted deployments of the same checkpoint are
+immune because they scope stops to content.
+
+Apply **[Patch 5](patches/0005-suppress-stops-in-reasoning.patch)** — bind-mount, no rebuild:
+
+```bash
+-v /path/to/patched/detokenizer.py:/opt/env/lib/python3.12/site-packages/vllm/v1/engine/detokenizer.py:ro
+```
+
+> **Both nodes.** `start-deepseek-v4-flash-dspark.sh` syncs the compose and env files to the
+> worker but **not** bind-mounted patch files. The file must exist at the same path on the
+> worker too, or it silently runs unpatched and you get confusing half-fixed results.
+
+Guard is per-request and needs no configuration: if the request's last prompt token is
+`<think>`, stop strings stay dormant until `</think>` appears. EOS and `max_tokens` are
+unaffected; non-thinking requests are untouched. Opt out with
+`VLLM_SUPPRESS_STOPS_IN_REASONING=0`.
+
+Measured on 2× DGX Spark (GB10), TP=2, k=5, `unsloth/DeepSeek-V4-Flash-0731`:
+
+| metric | before | after |
+|---|---|---|
+| decapitation reproducer (seed + stop) | 43 tok, `content: null` | 344 tok, correct answer |
+| stop honored on a non-thinking request | ✓ | ✓ (cuts at exact stop) |
+| GSM8K n=50, temp 0.6 / top_p 0.95 | 8–15 nulls, 0.66–0.84 | **1 null, 0.98** |
+
+The single residual null is mechanism (B) in [#18](../../issues/18) — a marginal-stability
+reasoning runaway, unrelated to stop strings. Patch 5 does not address it.
+
 Apply **[Patch 4](patches/0004-dspark-shared-expert-gate-up-proj.patch)** on top of your existing
 setup. Two lines, no rebuild — bind-mount it read-only:
 
@@ -261,6 +299,7 @@ image, same patches, same `k=5`, same `nvfp4_ds_mla` KV cache, same 1M context, 
 | Safety refusals | stock | removed (reported 32/32 bypass on a hard refusal suite) |
 | DSpark MTP draft modules | stock | **stock, unedited** |
 | Patch 4 | required | required |
+| Patch 5 | recommended (any stop-sending harness) | recommended |
 | On-disk | ~156 GB | ~156 GB |
 
 **Getting the uncensored weights:**
--- a/patches/0005-suppress-stops-in-reasoning.patch
+++ b/patches/0005-suppress-stops-in-reasoning.patch
@@ -18,22 +18,53 @@ WHY
 
 WHAT
   Per-request guard, no configuration required: at detokenizer construction, if
-  the request's LAST PROMPT TOKEN is `<think>`, stop strings stay dormant until
-  `</think>` appears in the output. EOS and max_tokens are unaffected.
-  Non-thinking requests are untouched by construction.
-  Opt-out: VLLM_SUPPRESS_STOPS_IN_REASONING=0
+  the request's LAST PROMPT TOKEN is the reasoning start marker, stop strings
+  stay dormant until the reasoning end marker appears in the output. EOS and
+  max_tokens are unaffected. Non-thinking requests are untouched by
+  construction.
+
+  The markers come from --reasoning-config (reasoning_start_str /
+  reasoning_end_str) when the deployment sets them, falling back to
+  <think> / </think>. The guard is therefore not DeepSeek-specific.
+
+  Opt-out: VLLM_SUPPRESS_STOPS_IN_REASONING=0. NOTE this is process-wide, not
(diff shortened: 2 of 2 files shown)


tonyd2wild:
Verified end to end on the maintainer rig and it holds up. Details, then a short change list.

**The bug is real here.** Live on the production endpoint (2× DGX Spark, TP=2, `0.21.1rc1.dev339+g1967a5627bc3`, k=5, `nvfp4_ds_mla`, temp 0, unpatched):

| request | finish | tokens | content | reasoning |
|---|---|---|---|---|
| `thinking:true`, `stop:["="]` | stop | 27 | **null** | cut at `47*53 `, right before the `=` |
| `thinking:false`, `stop:["="]` | stop | 41 | 154 chars, cut at the `=` (correct) | null |
| `thinking:true`, GSM8K, `stop:["Question:"]` | stop | 19 | **null** | cut at `User asks "` |

First attempt, deterministic. Your diagnosis is confirmed.

**Your precondition holds, and checking it corrected a note of ours.** In this checkpoint's tokenizer `<think>` is token **128821** and `</think>` is **128822** — single ids, and *not* in `added_tokens_decoder`, so `skip_special_tokens=True` doesn't strip them from `output_text`. That's precisely why the guard can latch on `</think>`; had they been special tokens it would have stayed armed forever. (We'd previously written that these were "plain strings, not added tokens" — half right, and the half that was wrong matters here.) Rendering the prompt for all three `reasoning_effort` values: thinking mode ends `<｜Assistant｜><think>` (128821), chat mode ends `<｜Assistant｜></think>` (128822). So the guard arms in thinking mode and provably cannot arm in chat mode.

**Applies and works.** `patch -p1 --dry-run` against the image's `detokenizer.py` is clean and the result parses. A/B in throwaway containers, feeding 6 tokens per `update()` to mimic k=5:

| arm | unpatched | patched |
|---|---|---|
| thinking prompt + `stop:["Question:"]` | stops at 12 tok, `</think>` never reached | guard on, runs to 28 tok, `</think>` present, stop fires in **content** at the right place |
| chat prompt + same stop | stops at 12 tok | guard off, byte-identical |
| thinking prompt, no stops | unchanged | unchanged |

### Changes I'd like before merge

1. **One-chunk leak.** With spec decoding `update()` gets up to k+1 tokens at once. The chunk containing `</think>` sets `_reasoning_closed = True` and *then* evaluates stops over `new_char_count = len(output_text) - stop_check_offset`, which covers the whole chunk — including the reasoning that preceded `</think>` in that same chunk. A stop string in the last ≤k tokens of reasoning can still fire. Please advance `stop_check_offset` past the `</think>` match before calling `check_stop_strings`.
2. **Document the interaction with #18(B).** If `</think>` never arrives, stops stay dormant for the whole request, so a runaway now runs to `max_tokens` where a stop might previously have cut it short. Correct by design, but it should be one line in the README so nobody reads it as a regression.
3. **Drop the bare `except Exception: pass`** — a renamed attribute would make this a silent no-op, and the failure mode is "the fix quietly isn't there". `logger.debug` on the exception is enough.

Nits, take or leave: source the markers from `--reasoning-config` (`reasoning_start_str`/`reasoning_end_str`) rather than hardcoding, since this deployment already passes them explicitly and it makes the generality claim in the header real; key the compatibility note on the vLLM build string `0.21.1rc1.dev339+g1967a5627bc3` rather than the image tag (current prod image is `mia-raf-pr1-nvfp4-probe-c-keys-concurrency-p2b` and it applies cleanly there); give the README section its own `##` rather than sitting under `### What to change`, where it reads as part of the Patch 4 story; hoist `import os`; and note that the opt-out is process-wide, so a caller who deliberately wants to bound thinking with a stop string has no per-request escape.

With 1–3 in, this is a merge from my side. Thank you for dogfooding it for four days first — and for the clean separation between mechanism (A) and (B) over in #18. That distinction is doing a lot of work.

Capicua25x:
Thanks for the review — all three are in, plus the nits. `31804a1`.

## 1. One-chunk leak

You were right, and my first attempt to verify it was wrong in an instructive way. Fixed by advancing `stop_check_offset` past the end marker when the guard closes, so stops are evaluated only over what follows it.

A/B against the currently-deployed version, real classes and real tokenizer, **one `update()` carrying 11 tokens with the stop and `</think>` both inside**:

| | stop fired | resulting `output_text` |
|---|---|---|
| before | `'Question:'` | `'restating '` — decapitated mid-reasoning |
| after | `None` | `'restating Question:</think>The answer is 4.'` |

Controls, identical on both: a stop present in **content** still fires (`'STOPHERE'`); a chat-mode prompt (last prompt token `</think>`) never arms the guard and fires exactly as unpatched.

**Worth flagging how nearly I shipped this unverified.** My first integration test chunked the input at 6 tokens/`update()` and reported "no stop fired" — which I read as the fix working. Then I ran the same test against the *old* file and got an identical result: at that chunking the stop landed in an earlier chunk than the close, so the test never exercised the leak at all. It only appears when both arrive in the same call, which is exactly what spec decoding produces and my test wasn't producing. Green result, code path never executed.

## 2. Interaction with #18(B)

Documented in the README under its own `##`. I went slightly past the ask, because there's a measurable consequence worth stating: since stops now stay dormant for a request where `</think>` never arrives, (B) runs to `max_tokens` where a client stop could previously truncate it by accident. So a fleet applying this may see *reported* token usage rise on those requests. The failure rate is unchanged — only how long each failure takes to admit it. Explicitly not a regression, and it removes stop strings as a confound in either direction when measuring (B).

## 3. Bare `except`

Now `logger.debug`. Verified by forcing the failure you were worried about — a tokenizer that raises where the guard reads it:

```
request survives: True · guard armed: False
DEBUG: 'stop-in-reasoning: guard not armed (simulated: renamed attribute)'
```

Request completes normally, guard stays disarmed, and it leaves a trace instead of vanishing.

## Nits taken

- **Markers from `--reasoning-config`** (`reasoning_start_str` / `reasoning_end_str`), falling back to `<think>` / `</think>`. Verified it actually *reads* them rather than merely not crashing: with a config carrying `<RAZONA>`/`</RAZONA>` it returns those; with none, the fallback. The generality claim in the header is now real.
- **Compatibility keyed on the build string** `0.21.1rc1.dev339+g1967a5627bc3` rather than an image tag, and your `mia-raf-pr1-...` result is noted alongside ours.
- `import os` hoisted to module scope.
- Opt-out documented as **process-wide, not per-request** — a caller who deliberately wants to bound reasoning with a stop string has no escape hatch short of a restart. Worth someone objecting to if they need one.
- README section given its own `##` rather than sitting under *What to change*.

## Two limits on the validation, stated rather than buried

- The `request` object in these tests is a stub. It carries every field the detokenizer touches, but real `EngineCoreRequest` construction isn't exercised.
- Marker sourcing is proven against an injected config, not against a live server started with a non-default `--reasoning-config`. Our deployment passes the defaults explicitly, so it's the fallback value here either way.

## On your `reasoning_effort: "none"` finding

That one lands on us: it's a fourth null-content generator and it invalidates any arm using `"none"` as a thinking-off control. Ours used `chat_template_kwargs {"thinking": false}`, which looks clean, but we're auditing before we report anything further on (B) rather than assuming.

Separately, and only relevant to (B), so I'll keep it short here and post properly in #18: we ran the same item across ~20 independent stacks serving this checkpoint via OpenRouter, provider pinned, cap above the vendor's observed maximum. Roughly **half of professional stacks fail to terminate on it**, and the disclosed quantization doesn't predict which. Happy to retitle-and-scope #18 to (B) as you suggested — that result belongs there rather than here.

Capicua25x:
One more data point that changes what this patch is, found while checking whether it had a shelf life.

## The bug is not specific to this runtime — it is in upstream vLLM today

I wanted to know whether a newer runtime had already fixed this, so that the compatibility note wouldn't outlive the defect. It hasn't. Same reproducer, version-agnostic (it touches nothing of the patch — no attributes, no env var, just "does a client stop string fired inside reasoning decapitate the request"):

| runtime | stop fired inside reasoning | resulting `output_text` |
|---|---|---|
| `0.21.1rc1.dev339+g1967a5627bc3` (this recipe, unpatched) | `'Question:'` | `'User asks "'` |
| `0.25.2.dev0+g752a3a504` (`ghcr.io/anemll/dspark-vllm-gx10:0.1.1`) | `'Question:'` | `'User asks "'` |

Identical truncation point, four minor versions apart. Control on both: a chat-mode prompt (last prompt token `</think>`) fires the stop normally, as expected.

And it is not something either build introduced. Upstream `vllm-project/vllm` `main`, fetched from GitHub just now:

```python
# 2) Evaluate stop strings.
stop_string = None
if self.stop and self.num_output_tokens() > self.min_tokens:
    stop = check_stop_strings(
        output_text=self.output_text,
        new_char_count=len(self.output_text) - stop_check_offset,
        stop=self.stop,
```

That is byte-for-byte the code this patch modifies: stop strings evaluated against the whole output stream with no notion of whether generation is currently inside the reasoning segment. I also checked that this recipe's overlay contains no `detokenizer.py` and none of its patches touch that file, so the copies in both images are what the respective vLLM builds produced from upstream source.

## What that means for scope

This is not a workaround for one pinned runtime. On current vLLM it affects **any think-in-prompt model** — DeepSeek V4, Qwen3, anything whose template ends the prompt with the reasoning-open token — hit by **any client that sends stop strings**, which includes lm-evaluation-harness on every request via `stop[:4]`. The failure is silent: HTTP 200, `finish_reason: "stop"`, `content: null`.

Two consequences you may want to weigh:

1. **The `APPLIES TO` note in the patch header is narrower than the defect.** Keyed to the build string it is accurate for this recipe, but a reader could infer the bug is version-specific. Happy to reword if you'd rather it say so explicitly.
2. **This probably belongs upstream.** If you have no objection I will open it against `vllm-project/vllm` with the reproducer and credit this thread — the guard as written is already sourced from `--reasoning-config` rather than hardcoded markers, so it is not DeepSeek-specific. I would rather ask than surprise you, since the diagnosis came out of your issue and your review caught the one-chunk leak in it.

## Unrelated, since it may save you a step

Your `0.25.2` bake-off result held up incidentally: we pulled that image only to run the test above, and the vLLM install sits at `/usr/local/lib/python3.12/dist-packages/` rather than `/opt/env/lib/python3.12/site-packages/`. So the bind-mount path in the README would need changing for that lineage — and mounting to the wrong path fails silently, exactly like the two-node worker case already documented in the header.

tonyd2wild:
All three are in, verified in the head diff: `stop_check_offset` advanced past the end marker on close, `logger.debug` in place of the bare except, and markers sourced from `--reasoning-config` with a `<think>`/`</think>` fallback. Merging.

The part of your follow-up I want to single out is the one you volunteered: your first integration test chunked at 6 tokens per `update()`, reported "no stop fired", and you read that as the fix working — until you ran it against the *old* file and got the same result. Green result, code path never executed. That's the failure mode that quietly ships broken patches, and you caught it on yourself. The 11-token single-`update()` A/B is the test that actually exercises the leak.

Your two stated limitations — stub `request` object, marker sourcing proven against an injected config rather than a live non-default `--reasoning-config` — are noted and acceptable. Stating them beats burying them.

**On taking it upstream: yes, please do**, and thank you for asking rather than surprising me. Your version-agnostic reproducer shows the identical truncation on `0.21.1rc1.dev339` and `0.25.2.dev0+g752a3a504`, and byte-for-byte the same code on `vllm-project/vllm` main. This affects any think-in-prompt model hit by any stop-sending client — lm-eval sends `stop[:4]` on every request — and the failure is silent: HTTP 200, `finish_reason: "stop"`, `content: null`. Your three-stack scoreboard makes it an ecosystem defect, not a recipe workaround.

Your `0.25.2` path observation is right and I owe you for it — that lineage installs to `/usr/local/lib/python3.12/dist-packages/`, not `/opt/env/...`. `check-patch3.sh` now resolves the package root instead of assuming ours (PR #28), since mounting to the wrong path fails silently — the same trap.

Issue #16, opened 2026-08-02
missing <think> token when generate the reasoning content


doongkc:
hello, I succesfully to build and run the model. 
But I'm facing to issue that the reasoning content not have <think> token when start the reasoning , only have </think> at the end. 

<img width="1280" height="97" alt="Image" src="https://github.com/user-attachments/assets/176352bf-d50e-4650-bc90-61a3a62cc018" />

I try a workaround in this post https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash-0731/discussions/26 but it didn't work. 

Just confirm that the think levels is worked ok, but the server response reasoning content as normal content, so the client show everything in the chat UI.
Do anyone have solution for this?

todoriri:
> hello, I succesfully to build and run the model. But I'm facing to issue that the reasoning content not have token when start the reasoning , only have at the end.
> 
[...]

Issue #8, opened 2026-07-09
Engine death under Hermes chat traffic: CUBLAS_STATUS_INTERNAL_ERROR -> EngineDeadError -> NCCL timeout, container exits 0


DaveCharland:
We hit a production outage on the Tony DeepSeek V4 Flash DSpark 2x DGX Spark recipe. This appears separate from #6: #6 is a soft empty/thinking-only parsed-content issue where the engine survives; this incident killed the vLLM engine and left the API down.

## Environment

- Repo: `tonyd2wild/DeepSeek-v4-Flash-DSpark-1M-NVFP4-KV-2x-DGX-Spark`
- Commit: `748f0346f92f8bb2de6f97bf49bfd3d8cc193c9a`
- Image tag: `vllm-dspark-runtime:dspark-nvfp4-stage-c-20260708`
- vLLM: `0.21.1rc1.dev339+g1967a5627bc3`
- Hardware: 2x DGX Spark / GB10, TP=2
- Model: `deepseek-ai/DeepSeek-V4-Flash-DSpark`
- Served model: `deepseek-v4-flash-dspark`
- API endpoint: `:8888`
- `max_model_len`: `1048576`
- `max_num_seqs`: `12`
- `kv_cache_dtype`: `nvfp4_ds_mla`
- `gpu_memory_utilization`: `0.85`
- async scheduling: enabled
- chunked prefill: enabled
- `max_cudagraph_capture_size`: `48`
- speculative config: `{"method":"dspark","num_speculative_tokens":3,"draft_sample_method":"probabilistic"}`

## Patch state

The committed bind-mounted `recipe/vllm/v1/spec_decode/dspark_proposer.py` crashed on this built image with `propose() got an unexpected keyword argument 'req_ids'`, so we followed your README guidance:

1. extracted `vllm/v1/spec_decode/dspark_proposer.py` from the built image,
2. replaced the bind-mounted proposer with that version,
3. reapplied `scripts/apply-nonuniform-guard.py`,
4. verified it compiled,
5. restarted and validated.

So this is not byte-for-byte the committed proposer file; it is the image-matched proposer with the non-uniform guard applied.

## What passed before the outage

After the proposer fix:

- `/v1/models` returned `deepseek-v4-flash-dspark`
- exact-response direct chat smoke passed
- short answer direct smoke passed
[...]

Issue #6, opened 2026-07-08
Soft-failure: completions with real tokens but empty/thinking-only parsed content (episodic, engine survives)


tonyd2wild:
Promoting @DaveCharland's report out of closed #2, per @paulbrav's suggestion — it's a distinct phenomenon and deserves its own tracking.

## Symptom (Dave's data, 2x Spark TP=2, stage-c-slotclamp + PR4)
- Completions return 55–433 real output tokens that parse to **no visible content** (sometimes thinking-only, sometimes neither content nor reasoning). HTTP 200, engine alive, no assert.
- ~13–25 soft events/day under agent load; ~1 hard episode/day: **deterministic for ~1–2 min on an identical prompt**, with a spec-acceptance collapse fingerprint (86%→14%, per-position craters), then full recovery. Full anatomy + table in Dave's 2026-07-06 comment on #2.
- NOT context-length-bound (40K/119K/180K episodes), NOT KV saturation.

## What we know (paulbrav's producer-side instrumentation, in #2)
- Persistent slot-index buffer host-visibly clean across crash repros — corruption materializes in **graph-replay-private state**.
- 12h compute-sanitizer across the recipe's Triton kernels: zero invalid accesses.
- **Acceptance collapse ≠ empty content**: with the slot clamp, stale-slot episodes explain the collapse (drafts transiently read the wrong request's KV → mass rejection → self-heal — predicted before Dave's table matched it). Rejected drafts cannot corrupt visible output, so the empty content needs a second mechanism.

## Candidate mechanisms for the empty content
(a) **Target-side graph-replay corruption** — in-range-but-wrong topk indices in the target's sparse indexer → garbled output that parses to nothing (fits 40K episodes; merged guards only bound OOB).
(b) **Reasoning-channel leakage under thinking:false** — parser eats a leaked reasoning stream. Supported: paulbrav runs thinking:true and has never seen this mode; Dave's THINKING_ONLY calls. Unexplained: the no-content-no-reasoning calls.
(c) **Tool-call parser swallowing** *(new, ours)* — Dave's episode call #4 was 'EMPTY **after tool calls**'; with --enable-auto-tool-choice, a malformed/unterminated tool-call open tag during degraded generation swallows the entire remainder — explaining exactly the 'neither content nor reasoning' calls that (b) cannot. Testable at the parser level without a GPU.

[...]

Issue #22, opened 2026-08-10
k=7 boots and generates on the anemll 0.25.2 image (768K/seqs-12) — and the two k>5 claims in this repo disagree with each other


robotnurse:
**Data point: `MTP_NUM_TOKENS=7` boots *and generates correctly* on the anemll 0.25.2 image at 768K / `max-num-seqs 12`. Flagging it because the two statements in this repo about k>5 also disagree with each other — and because the difference may be runtime-specific rather than an error.**

We are **not** advocating k=7. Our own acceptance data says it is not worth taking, and we run k=3 in production. This is only about the stated failure mechanism.

### The two claims

- `8a62e8b` — k>5 boots, but *"generation then fails because the drafter emits exactly `dspark_block_size` tokens per pass"*
- `05e56ee` — k=7 is *"rejected at boot — must divide evenly by `n_predict=5`"*, and k is *"effectively locked at 5"*

One says it boots and then fails at generation; the other says it never boots. Both cannot describe the same build.

### What we measured

2× DGX Spark GB10, TP=2, `deepseek-ai/DeepSeek-V4-Flash-0731` (snapshot `9e165c30`),
**`ghcr.io/anemll/dspark-vllm-gx10:0.1.1`** (vLLM `0.25.2.dev0+g752a3a504`), `max-model-len 768000`, `max-num-seqs 12`, `nvfp4_ds_mla`, `draft_sample_method: probabilistic`.

Booted healthy, `num_spec_tokens=7`, and generated correctly:

| prompt | finish | out | content |
|---|---|---|---|
| `What is 47*53?` | stop | 3 | `2491` ✅ |
| `Write exactly three words.` | stop | 5 | `Exactly three words.` ✅ |
| `Count from 1 to 20` | stop | 59 | `1, 2, 3, 4, 5, …` ✅ |

Zero engine errors. (An initial grep of ours reported 2 "errors" — both were the pattern matching `size_asserts` / `alignment_asserts` inside the `inductor_compile_config` dump. Not failures. Mentioning it so nobody repeats the mistake.)

`config.json` on this checkpoint does have `dspark_block_size: 5`, so the divisibility premise is there — it just did not produce either described failure here.

### Why we still declined k=7

Acceptance counters from `/metrics` after the generation test — 31 drafts, 217 draft tokens, 99 accepted:

[...]

PR #23, opened 2026-08-10
Fix four recipe defects surfaced by issues #6, #8 and #16

tonyd2wild:
Four verified defects, each traced to a specific issue thread. No behaviour change beyond what is described.

**1. `docker-compose.dspark.yml`: remove `--max-cudagraph-capture-size`** (issue #8)
`DEFAULT-CONFIG.md:87` says leave it unset; the compose still computed `MAX_NUM_SEQS × (MTP_NUM_TOKENS + 1)`. At `MAX_NUM_SEQS=1` that yields **6** — the exact value that logged `Truncating max_cudagraph_capture_size to 4` in @frankfurth's report, seven minutes before his worker died. Invisible at the shipped defaults (6 × 6 = 36), which is why it survived.

**2. `docker-compose.dspark.yml`: add `restart: unless-stopped`** (issue #8)
Promised in that thread, never landed. The container exits 0 on `EngineDeadError`, so nothing restarts it.

**3. `benchmarks/soak.py`, `benchmarks/garble_tap.py`: read `reasoning`** (issues #6, #16)
This runtime returns `message.reasoning` / `delta.reasoning`. `reasoning_content` is deprecated and accepted only on input — there is no such key in a response. Both harnesses read only the old name, so any thinking-on run emits false `EMPTY`/`SHORT-BODY` events and `garble_tap`'s streaming path drops reasoning entirely. Now prefers `reasoning` with a fallback. Does not invalidate the 553-request soak (ran `thinking:false`).

**4. `README.md`: new 'Reasoning / thinking mode' section** (issue #16)
The repo documented none of this. Covers the field name, that `<think>` is written into the *prompt* so a missing opening tag is correct behaviour, that `reasoning_effort:"none"` + `thinking:true` returns `content: null` (measured 4/4 — a false-positive generator for #6), that `low` behaves as `high` on the stock build, that `chat_template.jinja` is ignored under `--tokenizer-mode deepseek_v4`, and that thinking mode needs output budget.

Opened as a PR rather than pushed to main so the wording can be reviewed.

### DIFF
--- a/DSPARK-SHARED-EXPERT-FIX.md
+++ b/DSPARK-SHARED-EXPERT-FIX.md
@@ -231,8 +231,10 @@ Two things previously documented here were wrong, and are fixed in this revision
 
 2. **The divisibility rule was stated imprecisely.** The guard only fires when
    `k > n_predict`, which is why `k=3` and `k=4` boot fine. The accurate rule is: **`k ≤ 5`, or a
-   multiple of 5.** Note upstream vLLM uses a `>=` check for DSpark instead, so upstream permits
-   `k=7` at config time — and then hits the same tensor-shape crash.
+   multiple of 5.** Upstream 0.25.2 uses the **same** `>` check, but resolves `n_predict` to
+   `num_nextn_predict_layers` (1 for this checkpoint) rather than `dspark_block_size`, so the
+   guard never fires there; and its DSpark speculator sizes the draft block from `k`, so no shape
+   mismatch follows. See issue #22.
 
 Where k=7 *does* work is on drafters whose draft block is at least 7 — e.g. MiMo-V2.5 DFlash
 (`block_size` = 8, run here at `num_speculative_tokens: 7`), GLM-5.2's DSpark speculator
--- a/README.md
+++ b/README.md
@@ -99,13 +99,29 @@ therefore measures **steps/s, not tokens/s**, and under-reports by the acceptanc
 vs 60.1 tok/s on the identical request. Read `usage.completion_tokens`, or divide server-side
 `vllm:generation_tokens_total` by wall time.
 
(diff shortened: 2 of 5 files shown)


tonyd2wild:
Removed the NVFP4 KV caveat from this PR. The reading behind it (`get_kv_cache_shape` returning 584 for both dtypes on V4, and stage-c rolling 416 → 584) is what the allocation code says, but it is disputed and code-reading alone doesn't settle it. The decisive test is a same-node, same-gmu A/B on the reported `GPU KV cache size: N tokens` between `nvfp4_ds_mla` and `fp8_ds_mla` — not publishing a claim about the repo's headline feature ahead of that number.

Also corrected in this branch: the weights are **FP8 e4m3**, confirmed by arithmetic rather than inference — 165,265,454,782 elements in 155.4 GiB = **1.010 bytes/param (8.08 bits)**, with 148B of those elements sitting on disk as I8 at a full byte each. An earlier read of mine treated `expert_dtype resolved to 'fp4'` in the boot log as evidence of 4-bit weight storage; that log line describes the MoE compute backend, not the on-disk dtype.

What remains here is verified and independent of any of that:

1. `docker-compose.dspark.yml`: remove `--max-cudagraph-capture-size` (contradicted `DEFAULT-CONFIG.md:87`; at `MAX_NUM_SEQS=1` it produced exactly the value that truncated to 4 in #8)
2. `docker-compose.dspark.yml`: add `restart: unless-stopped` (promised in #8, never landed)
3. `benchmarks/soak.py` + `garble_tap.py`: read `reasoning`, not `reasoning_content`
4. `README.md`: new Reasoning / thinking mode section
5. `README.md` + `DSPARK-SHARED-EXPERT-FIX.md`: k>5 is a property of the runtime, not the checkpoint (#22), and strike the false claim that upstream uses a `>=` check — it uses the identical `>` check with a different `n_predict`

PR #24, opened 2026-08-10
Reasoning effort and script fixes

hhackbarth:
Split out of #17 so the tool-parser change can be reviewed on its own. This half contains no parser code. Requested in https://github.com/tonyd2wild/DeepSeek-v4-Flash-0731-DSpark-1M-NVFP4-KV-2x-DGX-Spark/pull/17#issuecomment

reasoning_effort — three levels restored. This fork's base image ships one constant, REASONING_EFFORT_MAX, byte-identical to DeepSeek's high text (476 chars), plus a normaliser that collapses everything except max/xhigh to "high" — a level that then injects nothing. Net effect: low doesn't exist as a distinct level, and DeepSeek's real max (526 chars) is unreachable. Verified against ghcr.io/bjk110/vllm-spark:unholy-fusion-prod-ready, quoted in the Dockerfile comment so the next reader doesn't have to guess which image was inspected. The replacement strings are extracted programmatically from the checkpoint's own encoding/encoding_dsv4.py rather than retyped.

⚠️ Behaviour change: reasoning_effort="max" now injects DeepSeek's max text instead of its high text — existing callers of "max" get a materially stronger instruction. "low" and omitted render byte-identically to before; verified live via /tokenize (prompt lengths 51 / 527 / 577 for low / high / max).

DEFAULT_CHAT_TEMPLATE_KWARGS in compose defaults to the previous hardcoded '{"thinking":false}', so nothing changes unless it's set. It exists so the restored levels are reachable from configuration rather than only per request. Happy to drop it if you'd rather that land separately.

Port derivation. start-/status-/smoke- set their health-check URLs before sourcing .env.dspark, pinning them to 8888. A non-default VLLM_PORT therefore starts the server correctly and then times out against the wrong port — a confusing failure, since the server is healthy the whole time. URLs are now derived after the source; an explicit API_URL still wins.

agent_sanity_bench. Adds an empty-output check, counted separately and reflected in the exit code, plus reasoning_chars per row. Empty text trips none of the existing CJK/repeat/XML checks, so the gate passed with exit 0 while most responses carried no visible content.

⚠️ This now fails closed: with thinking on at the default MAX_TOKENS=256 it reports failures that previously passed silently. Verified in both directions — exit 1 at 256, pass at an adequate budget.

Ordering: this and #17 both touch the py_compile list in the same Dockerfile block. Merge #17 first as planned and I'll rebase this one.

### DIFF
--- a/docker-compose.dspark.yml
+++ b/docker-compose.dspark.yml
@@ -126,6 +126,11 @@ services:
       MASTER_PORT: "${MASTER_PORT:-25000}"
       MTP_NUM_TOKENS: "${MTP_NUM_TOKENS:-5}"
 
+      # Server-side chat-template defaults. Optional; empty keeps the historical
+      # '{"thinking":false}'. This is what makes the restored three-level
+      # reasoning_effort reachable from configuration instead of per request.
+      DEFAULT_CHAT_TEMPLATE_KWARGS: "${DEFAULT_CHAT_TEMPLATE_KWARGS:-}"
+
     command:
       - bash
       - -lc
@@ -136,6 +141,9 @@ services:
         export CUDAToolkit_ROOT="$${CUDAToolkit_ROOT:-$${CUDA_HOME}}";
         export LD_LIBRARY_PATH="/opt/env/lib:/opt/env/targets/sbsa-linux/lib:$${LD_LIBRARY_PATH:-}";
         SPECULATIVE_CONFIG="{\"method\":\"dspark\",\"num_speculative_tokens\":${MTP_NUM_TOKENS:-5},\"draft_sample_method\":\"probabilistic\"}";
+        CHAT_TEMPLATE_KWARGS="$${DEFAULT_CHAT_TEMPLATE_KWARGS:-}";
+        [ -n "$${CHAT_TEMPLATE_KWARGS}" ] || CHAT_TEMPLATE_KWARGS='{"thinking":false}';
+        echo "serving with chat-template-kwargs=$${CHAT_TEMPLATE_KWARGS}";
         exec /opt/env/bin/vllm serve ${DSPARK_MODEL:-deepseek-ai/DeepSeek-V4-Flash-DSpark}
         --served-model-name ${SERVED_MODEL_NAME:-deepseek-v4-flash-dspark}
         --host ${VLLM_HOST:-127.0.0.1}
@@ -159,7 +167,7 @@ services:
         --enable-auto-tool-choice
         --reasoning-parser deepseek_v4
         --reasoning-config '{"reasoning_parser":"deepseek_v4","reasoning_start_str":"<think>","reasoning_end_str":"</think>"}'
-        --default-chat-template-kwargs '{"thinking":false}'
+        --default-chat-template-kwargs "$${CHAT_TEMPLATE_KWARGS}"
         --generation-config vllm
         --enable-flashinfer-autotune
         --nnodes 2
--- a/recipe/Dockerfile.dspark-runtime-overlay
+++ b/recipe/Dockerfile.dspark-runtime-overlay
@@ -25,6 +25,33 @@ COPY vllm/v1/attention/backends/mla/b12x_mla_sparse.py /opt/env/lib/python3.12/s
 COPY vllm/v1/spec_decode/dspark.py /opt/env/lib/python3.12/site-packages/vllm/v1/spec_decode/dspark.py
 COPY vllm/v1/spec_decode/dspark_proposer.py /opt/env/lib/python3.12/site-packages/vllm/v1/spec_decode/dspark_proposer.py
 COPY vllm/v1/worker/gpu_model_runner.py /opt/env/lib/python3.12/site-packages/vllm/v1/worker/gpu_model_runner.py
+# reasoning_effort three-level fix.
+#
+# Verified against THIS fork's base image, ghcr.io/bjk110/vllm-spark:
+# unholy-fusion-prod-ready -- not against upstream vLLM main, whose tokenizer
+# differs. Two facts from that image:
+#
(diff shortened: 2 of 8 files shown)


tonyd2wild:
Close. One real bug, then I'll merge.

**`agent_sanity_bench.py` reads the wrong field.**

```python
"reasoning_chars": len(data["choices"][0]["message"].get("reasoning_content") or ""),
```

This runtime returns reasoning in **`reasoning`**; `reasoning_content` is deprecated and input-only, and no such key appears in a response. So `reasoning_chars` will be `0` on every row and the new column reports nothing. It's non-streaming here, so `message.reasoning`. Both existing harnesses already do `msg.get("reasoning") or msg.get("reasoning_content")` — matching them is the safest form. The `is_empty` docstring names the wrong field too.

Worth flagging that this is the exact bug the empty-output gate exists to catch, and that two of my own harnesses had it until last week.

**Everything else confirmed.** I executed your vendored `REASONING_EFFORT_PROMPTS`: `low` 0 chars, `high` 476, `max` 526 — and the 476-char string is byte-identical to what the stock build ships as its single `REASONING_EFFORT_MAX` constant. That settles #25: stock `"max"` really is DeepSeek's *high* text, and real `max` is unreachable. Both vendored files compile clean. The port-derivation fix is a genuine bug — today a non-default `VLLM_PORT` starts the server correctly and then times out against 8888.

The `reasoning_effort="max"` behaviour change is release-noted in your commit message and now in the README's *Reasoning / thinking mode* section too.

**Heads-up on rebase:** `docker-compose.dspark.yml` has moved — PR #28 adds a node-local JIT-cache block (issue #27) and `VLLM_ENGINE_READY_TIMEOUT_S`. Different regions from your `DEFAULT_CHAT_TEMPLATE_KWARGS` addition, so it should merge cleanly, but the file isn't where you left it. #17 also merged just now.

Fix the field name and I'll merge.

hhackbarth:
Fixed and rebased — `8bceae3`.

**You're right, and it's worse than a typo on my side**: I'd identified this exact field name a week ago while debugging why a client showed reasoning but no answer, wrote it down, and then queried `reasoning_content` anyway when adding the column. It would have reported `0` on every row - a silent zero in the check whose whole purpose is catching silent zeros. Now uses `message.get("reasoning") or message.get("reasoning_content")`, matching `benchmarks/garble_tap.py`, and the `is_empty` docstring names the right field.

Verified against the live server rather than by inspection:
```
row 0: reasoning_chars= 1650   content=200 ch
row 1: reasoning_chars= 4054   content=200 ch
```
**Rebased onto current main**. The direct rebase conflicted because this branch's history still carried the parser commit that #17 merged, so I rebuilt it on `upstream/main` and re-applied the two touched files by hand. Thanks for the heads-up on #28 - compose has the node-local JIT block and `VLLM_ENGINE_READY_TIMEOUT_S` now, and my `DEFAULT_CHAT_TEMPLATE_KWARGS` addition sits in a different region as you predicted. The Dockerfile already carries the parser and the import check from #17, so this only adds the two tokenizer `COPY` lines and their `py_compile` entries.

Thanks also for confirming the three-level strings independently - `low` 0 / `high` 476 / `max` 526, with the 476-char one matching the stock single constant. That's the same result I got extracting them from the checkpoint, and it settles #25.

hhackbarth:
Just a little reminder @tonyd2wild , that after my rebase this branch is still mergeable.
It would help me to return to the main branch also locally.

tonyd2wild (commented):
The `reasoning_content` fix lands exactly as asked — `message.get("reasoning") or message.get("reasoning_content")`, matching garble_tap.py's precedence, and the `is_empty` docstring names the right field. Verified the effort strings are 476/526 chars exactly, consistent with your /tokenize numbers (51+476=527, 51+526=577), and the low/high/max mapping is coherent including unknown-values-fall-to-low. Port derivation ordering is right in all three scripts and explicit `API_URL`/`CHAT_URL`/`PORT` overrides still win. The empty-output gate is a real gap-fix, and nothing in CI runs the bench, so the fail-closed exit only affects thinking-enabled manual runs — as intended.

One small ask before merge: you added the two tokenizer files to `py_compile` but not to the import-check `CHECKS` list, and they're exactly the kind of file that block exists for (relative imports of `.hf`/`.protocol`, plus the `regex` dependency, none of which py_compile sees). Please add `("vllm.tokenizers.deepseek_v4", "DeepseekV4Tokenizer")` and `("vllm.tokenizers.deepseek_v4_encoding", "encode_messages")` to `CHECKS` — or say the word and I'll add them in a follow-up commit on merge.

Two notes, no code change needed:

1. The body flags the `"max"` behavior change, but `"high"` also changes — stock `"high"` injected nothing, after this PR it injects the 476-char text. Callers already sending `"high"` get materially stronger instructions too. Going in the merge notes.
2. Before merging I'll diff `deepseek_v4_encoding.py` against the base image's copy to confirm the only delta is the effort table, since that claim isn't verifiable from the repo alone. That's a ten-minute check on our lane, not a request to you.

tonyd2wild:
Merged. The base-image gate passed cleanly: extracted `deepseek_v4_encoding.py` from `vllm-dspark-runtime`'s installed tree and diffed against your file — 30 diff lines, both hunks exactly the described change, nothing else touched in 765 lines. The diff also independently confirms your central claim: the stock constant named `REASONING_EFFORT_MAX` is byte-identical to your `high` text, so stock `max` really was serving DeepSeek's *high* prompt with the real max unreachable. Nice find, and thanks for the programmatic extraction — it showed.

The two tokenizer modules are now in the import-check `CHECKS` list as a follow-up commit on main, as offered. Merge notes carry the second behavior change we discussed (`high` goes from injecting nothing to injecting the 476-char text).

Issue #27, opened 2026-08-11
NFS-shared HF cache across nodes breaks all seven JIT caches — document or default per-node


antoniohlc:
First: thank you for this stack — we reproduced your numbers on 2× DGX Spark (direct CX7 link) with warm single-stream decode at 69.8–73.8 tok/s on the 0731 checkpoint, right inside your reported band. The claims are real.

But getting there cost us a full day on one root cause that I think is worth a README warning (or better, a compose change), because it will hit anyone whose HF cache is shared between the two nodes.

## The problem

The compose file mounts `HF_CACHE` at `/cache/huggingface` and sets

```
VLLM_CACHE_ROOT: /cache/huggingface/vllm-cache
```

which implicitly assumes that directory is **node-local**. On a 2-node setup it's natural to share the HF cache over NFS instead (download 167 GB once, serve on both nodes) — and then **seven** JIT/workspace caches end up inside the shared mount, because they all default to (or are placed under) the cache tree:

1. vLLM compile cache (`VLLM_CACHE_ROOT`)
2. DeepGEMM JIT (`DG_JIT_CACHE_DIR`)
3. TileLang (`TILELANG_CACHE_DIR`)
4. TorchInductor (`TORCHINDUCTOR_CACHE_DIR`)
5. Triton (`TRITON_CACHE_DIR`)
6. torch extensions (`TORCH_EXTENSIONS_DIR`)
7. FlashInfer workspace (`FLASHINFER_WORKSPACE_BASE`)

Both ranks then JIT into the same directories concurrently, and the failures are serial and misleading — each fix uncovers the next:

- a `torch.compile` `FileExistsError`/`makedirs` race at startup,
- DeepGEMM `runtime != nullptr` asserts from stale/half-written cubins read over NFS,
- and the killer: an ABI-mismatched FlashInfer `sampling.so` compiled by one node and silently loaded by the other — silent precisely because the stack runs `FLASHINFER_DISABLE_VERSION_CHECK=1`. That one downed the head's first engine attempt on every boot, which (with the worker entrypoint not retrying after a process death) also orphaned the worker each time.

None of the error messages point at the cache being shared; they read like broken kernels or a broken build.

## The fix that made the stack serve

Override all seven onto a node-local mount:

```yaml
volumes:
  - ${HOME}/.cache/vllm-dspark:/vllm-cache
environment:
[...]

Issue #25, opened 2026-08-10
Docs: reasoning_effort low/high inject NO prefix (direction inverted), and max injects the model's high text


Capicua25x:
**Doc correction: `low`/`high` behave as *no effort prefix*, not as `high` — and the prefix that `max` injects is the model's `high` text, not its `max` text.**

The README's *Reasoning / thinking mode* section says:

> `reasoning_effort: "low"` behaves as `"high"` on the stock build — only `"max"`/`"xhigh"` differ. See PR #17.

The observation is right and the direction is inverted, which matters because it changes what a caller should do. `low` and `high` both produce **no reasoning-effort prefix at all**. Anyone selecting `high` on this stack is running at the *vendor's* `low`, silently.

## Verify in three requests

Any prompt works; `prompt_tokens` is the witness. The prefix is 79 tokens, so it either shows up or it doesn't.

```bash
for E in low high max; do
  printf '%-5s ' "$E"
  curl -s http://localhost:8888/v1/chat/completions \
    -H 'Content-Type: application/json' \
    -d "{\"model\":\"deepseek-v4-flash\",
         \"messages\":[{\"role\":\"user\",\"content\":\"hi\"}],
         \"max_tokens\":1,
         \"reasoning_effort\":\"$E\"}" \
  | python3 -c 'import json,sys; print("prompt_tokens =", json.load(sys.stdin)["usage"]["prompt_tokens"])'
done
```

On this recipe (`0.21.1rc1.dev339+g1967a5627bc3`, `--tokenizer-mode deepseek_v4`, `thinking:true`):

```
low   prompt_tokens = 5
high  prompt_tokens = 5      <-- identical to low: no prefix
max   prompt_tokens = 84     <-- +79
```

Same shape on a longer prompt (HLE item, system + user): `low` 681, `high` 681, `max` 760.

For contrast, the same three against `api.deepseek.com`, same messages:

```
no field  760      <-- their default IS high
low       681
high      760      <-- +79, the prefix
max       773
```

So the vendor separates all three; this runtime separates two, and the pair it merges is `low`+`high` at *zero* prefix.

## Why, from the runtime's own source

`vllm/tokenizers/deepseek_v4.py`:

```python
elif reasoning_effort in ("max", "xhigh"):
    reasoning_effort = "max"
else:
    reasoning_effort = "high"          # low / medium / high all land here
[...]

Issue #12, opened 2026-07-18
Where is the 2026-07-16 docker image hosted?


Archmonger:
Your 2026-07-16 speed update mentions that you are using `vllm-dspark-runtime:mia-raf-pr1-nvfp4-probe-c-keys-concurrency-p2b`. Where is this image hosted? If it's a custom image you made, any chance you can publish it to the GitHub Container Registry? I can't seem to replicate your performance numbers using strictly what is documented in this repo.

tonyd2wild:
@Archmonger — thank you for this, and sorry for the friction. Your "I can't replicate your performance numbers using strictly what is documented in this repo" turned out to be **a real bug in this repo, not a mistake on your side**. Two separate problems; one is now fixed.

## 1. The speed gap: a stale default (fixed, please re-pull)

The shipped `.env.dspark.example` and `docker-compose.dspark.yml` were still handing out **`MTP_NUM_TOKENS=3`** — the 2026-07-03 greedy-draft garble-fix value — even though Patch 3 made `k=5` garble-safe with `draft_sample_method: probabilistic`, and `DEFAULT-CONFIG.md`, `README.md` and `UPSTREAM_V024_STATUS.md` all specify **5**.

`DEFAULT-CONFIG.md` states it plainly: **`num_speculative_tokens: 5` … +24% over 3.**

So anyone cloning this repo and running the defaults was benchmarking the slow path — exactly what you hit. **Fixed on `main`:**

- `.env.dspark.example`: `MTP_NUM_TOKENS=5` (+ corrected comment explaining the history)
- `docker-compose.dspark.yml`: `${MTP_NUM_TOKENS:-3}` → `:-5` — this also repairs the derived
  `--max-cudagraph-capture-size`, computed as `MAX_NUM_SEQS × (MTP_NUM_TOKENS + 1)`, which was
  therefore *also* sized for k=3
- `README.md`: headline corrected `~50-60` → **`~70 tok/s`** (our own measured number)

Please `git pull` and re-run. That alone should move you most of the way to the documented numbers.

## 2. The image: honest answer

`vllm-dspark-runtime:mia-raf-pr1-nvfp4-probe-c-keys-concurrency-p2b` is **a locally-built image, not hosted anywhere** — that was an oversight in how the speed update was written, and it's fair to call it out.

[...]

PR #28, opened 2026-08-12
Issue sweep: GHCR workflow, node-local JIT caches, check-patch3 portability, reasoning_effort correction

tonyd2wild:
Five commits, each closing a specific reported gap. No runtime behaviour changes to the serving path.

**1. `.github/workflows/publish-ghcr.yml` (#12)** — the repo had no `.github/` directory at all, which was the actual gap; not a credential problem, as @Archmonger correctly pointed out. Triggers on release/tag/dispatch, `permissions: {packages: write}` with the automatic `GITHUB_TOKEN`, builds via the same `build-dspark-vllm-runtime.sh` humans run. Targets a **self-hosted arm64 runner** — the image is GB10 `sm_121` and ~19 GB, so hosted x86 runners aren't a path. Fails closed pre-push on non-arm64 result, missing Patch 3, or missing Patch 4.

**2. `scripts/check-patch3.sh` portability (#22)** — no longer hardcodes `/opt/env/lib/python3.12/site-packages`. Asks the container's Python where `vllm` lives, falls back to probing known layouts (including `/usr/local/lib/python3.12/dist-packages` on the 0.25.2 lineage), accepts a `VLLM_ROOT` override, and prints the *resolved* path in the remediation hint — a bind mount to the wrong path fails silently.

**3. Node-local JIT caches (#27)** — a new `/vllm-cache` volume (`JIT_CACHE_DIR`, default `~/.cache/vllm-dspark`) with all seven JIT paths pointed at it, independent of `HF_CACHE`. Weights over NFS are fine; the JIT tree must be per-node. The FlashInfer case is why this is a shipped default rather than a doc note: `FLASHINFER_DISABLE_VERSION_CHECK=1` is set here for unrelated reasons and turns a cross-node ABI mismatch into a silent bad load. Also mirrors `VLLM_ENGINE_READY_TIMEOUT_S` into compose so the compose and sparkrun paths don't diverge (owed on #19).

**4. Heredoc contract in `recipe/nvfp4/Dockerfile.stage-*` (#14)** — sparkrun's `pre_exec` reconstructs the overlay by `sed`-ing the heredocs out of these files, so their formatting is load-bearing. Now documented in-file. The comments deliberately never repeat either marker, which would move sparkrun's extraction range; the extractor was re-run against all three stages and produces byte-identical output.

**5. Docs (#25)** — `reasoning_effort` direction corrected per @Capicua25x: `low`/`medium`/`high` inject **no** prefix, only `max`/`xhigh` do, and what they inject is the model's *high* text — so real `max` is unreachable on this tokenizer mode. Verified independently by executing PR #24's vendored table: low 0 / high 476 / max 526 chars, and the 476-char string is byte-identical to the stock build's single `REASONING_EFFORT_MAX`. Also adds the null-content triage table (`finish_reason` × `</think>`-present) and the NFS JIT-cache trap.

Checks: all YAML parses, all shell scripts pass `bash -n`, `check-patch3.sh` exercised against a mock docker across four container layouts with correct exit codes, sparkrun's heredoc extractor verified byte-identical on all three stages, 73 relative markdown links checked.

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -29,6 +29,15 @@ HF_CACHE=${HOME}/.cache/huggingface
 # Optional worker-local Hugging Face cache. Use this when the worker user/path
 # differs from the head node. If blank, the worker uses HF_CACHE.
 WORKER_HF_CACHE=
+# JIT / compile caches. MUST BE NODE-LOCAL on every node — never NFS, and never
+# inside a shared HF_CACHE. Sharing it makes both ranks JIT into the same
+# directories and you get, in order: a torch.compile makedirs race, DeepGEMM
+# "runtime != nullptr" from half-written cubins, and an ABI-mismatched
+# FlashInfer sampling.so built by one node and silently loaded by the other
+# (silent because FLASHINFER_DISABLE_VERSION_CHECK=1). None of the errors name
+# the cache. Credit @antoniohlc, issue #27. Default is node-local already;
+# only set this if you keep HF_CACHE on NFS and want the JIT tree elsewhere.
+#JIT_CACHE_DIR=${HOME}/.cache/vllm-dspark
 HF_HUB_OFFLINE=0
 TRANSFORMERS_OFFLINE=0
 HF_HUB_DISABLE_XET=1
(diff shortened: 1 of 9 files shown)


tonyd2wild:
Rebased onto current main. **Before the rebase this PR would have reverted #17 and #21** — it was branched before those merged, so its diff deleted `patches/0005-suppress-stops-in-reasoning.patch` (-197) and `recipe/overlay/vllm/tool_parsers/deepseekv32_tool_parser.py` (-321). Now it deletes nothing and both files survive. Worth flagging as the failure mode: a stale branch whose diff silently removes work merged after it was cut.

PR #29, opened 2026-08-12
scripts: loop_detector.py — distinguish reasoning loops from heavy tails (from #18)

Capicua25x:
The loop detector promised in #18 — tells a reasoning **loop** apart from a **heavy tail**, from the text alone, free.

Both failures look identical from outside (`finish_reason=length`, `content=null`), but they want opposite fixes: a heavy tail wants more `max_tokens`; a loop makes more budget pure waste. The discriminator is **novelty of word 8-grams per 4k-char window**: healthy reasoning stays novel to the end; a loop collapses below 2% and never recovers.

It also carries the measurement trap we fell into (documented in the docstring): the DS4 loop is *templated*, not verbatim — one of our captured runaways is 92% unique at 120-char block granularity and 96% recycled at phrase level. Any fixed-window block-uniqueness check reports a templated loop as fresh text. Count recycled n-grams or unique lines.

**Validated today before this PR** — three captured ~100k-token runaways from the (B) investigation plus a fresh healthy control:

| trace | verdict | onset |
|---|---|---|
| runaway capture 1 (~89k tok) | LOOP | ~7k |
| runaway capture 2 (~94k tok) | LOOP | ~13k |
| runaway capture 3 (~101k tok) | LOOP | ~30k |
| healthy terminating reasoning (~7.2k tok) | HEAVY TAIL (no collapse) | — |

Onsets match the interleaving-with-completion-lengths picture from #18 — it's a race, not a threshold.

Usage:
```
python3 scripts/loop_detector.py trace.txt [trace2.txt ...]
cat trace.txt | python3 scripts/loop_detector.py
```
Feed it the **reasoning** stream, not `content`. Pairs well with the acceptance-rate telltale from the spec-decode metrics log (loops pin acceptance near the max).

### DIFF
--- a/scripts/loop_detector.py
+++ b/scripts/loop_detector.py
@@ -14,8 +14,18 @@ WHAT IT MEASURES
   novelty(window) = fraction of word 8-grams in this window that have not
                     appeared anywhere earlier in the same trace.
 
-  A healthy trace keeps novelty high to the end. A looping trace collapses and
-  never recovers.
+  Windows are taken over the WORD SEQUENCE (700 words per window), not the raw
+  character stream. Char-stream windows truncate tokens at every boundary; for
+  short-period verbatim loops the 1-2 novel boundary fragments dominate the
+  window's small distinct-shingle population and hold novelty above threshold
+  -- a one-sentence verbatim loop (CJK or English) can plateau near 40% novelty
+  and read as HEAVY TAIL forever. Tokenize once, window over words.
+
+  A healthy trace keeps novelty high to the end. A looping trace collapses AND
+  NEVER RECOVERS -- and the verdict enforces both halves: after onset, novelty
+  must stay below threshold to the end of the trace. A collapse that recovers
+  (a long verbatim re-quote of an earlier code block, say) is reported as
+  transient repetition, not as a loop.
 
 WHY NOT BLOCK UNIQUENESS
 
@@ -40,6 +50,18 @@ CALIBRATION
   novelty collapses at token ~7.9k / ~13.8k / ~29.7k and never recovers --
   afterwards it oscillates between 0 and 0.6%, never above. It does not sit at
   exactly zero, so do not test for `== 0`; the threshold below is 2%.
+  Re-verified after the switch to word-sequence windows: onsets ~7.2k / ~13.7k
+  / ~30.2k (each within one window of the char-stream calibration), and two
+  genuine long heavy-tail controls -- real length-capped-but-healthy traces of
+  ~31k and ~34k tokens -- never trip the threshold in any window.
+
+KNOWN LIMITATIONS
+
+  A FRAGMENTED loop -- recycled fragments interleaved with enough novel filler
+  that no single window's novelty drops below threshold -- is not detected
+  (reported by @brianmswheart on the PR thread; real traces exist). That shape
+  needs a cumulative recycled-mass tier, not a window verdict; out of scope
+  here, deliberately, so this tool stays a one-screen instrument.
 
 USAGE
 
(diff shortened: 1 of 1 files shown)


brianmswheart:
Field report from a 2x DGX Spark GB10 deployment (vLLM 0.25.2.dev, 1M ctx): at 245-257k-token agent contexts we hit the (B)-class loops live — but they **fragment**. The model emits 1-8k chars of cycled action-sentences ("Let me do it. Let me run it. Emitting now."), a tool call interrupts, then the next completion loops again. Each fragment sits below the 3x4k-char windowed floor, so the detector as written never fires while the session visibly degrades (turns ran 19-42 minutes).

We added a second tier that catches the fragments: flag when any <=80-char normalized sentence repeats >=8 times AND repeats make up >=40% of the text's sentences. Validated against varied-prose and code-like controls with no false positives; catches the observed shapes down to ~1k chars. Happy to PR it here if wanted.

One trap we hit while building test controls, worth a note in the README: a "clean" control constructed by repeating sentences IS a loop by this criterion — controls must be genuinely varied or they read as false positives.

Related: since the client replays the whole transcript, each loop fragment in history raises the probability of the next (same snowball as HF discussion #58). Implementing #58's strip-loop-text-from-resubmitted-history mitigation as a gateway-side pre-call scrub (collapse consecutive/short-cycle duplicate sentences in assistant history only) measurably broke the escalation on live sessions without restarting them.

tonyd2wild (changes requested):
Thanks — this is a genuinely useful spin-out of #18, and the docstring's documentation of the block-uniqueness trap alone earns its place in `scripts/`. Stdlib-only, linear on 100k-token traces, verified. One substantive change requested before merge, one smaller one:

**1. Window over the word sequence, not the char stream.** I tested the script on synthetic traces and found that windowing the raw char stream truncates tokens at every 4k boundary. For loops with a short period, the 1–2 novel boundary fragments dominate the window's small distinct-shingle population and keep novelty above the 2% threshold: a verbatim Chinese loop (one sentence repeated — i.e., the #18 deterministic-reproducer shape, "same ~20-word fragment", CJK drift) plateaus at a steady 40% novelty and is classified HEAVY TAIL forever; a 21-word English verbatim loop starting at ~8k tokens isn't flagged until ~80k. Tokenizing the full text once and windowing over words (~700 words ≈ your 4k chars) fixes both in my tests and leaves the templated-loop cases you calibrated on unchanged. It also removes the `<8 words → novelty 0.0` dry-window edge case.

**2. Check that the collapse persists.** The docstring's discriminator is "collapses *and never recovers*", but `verdict()` returns at the first 3 dry windows. A healthy trace that verbatim re-quotes ~12–15k chars (a long code block restated) becomes a false LOOP. After onset, verify novelty stays below threshold to the end of the trace; otherwise report it as transient repetition. This also makes the "(novelty stays below X thereafter)" line always truthful.

Non-blocking notes: the validation table would be stronger with one genuine *long* heavy-tail control (length-capped but healthy) — the short terminating control doesn't exercise the boundary the tool exists to draw; and @brianmswheart's fragmented-loop report is a real second failure shape worth a "known limitations" line in the docstring. Happy to merge once (1) and (2) are in.

Capicua25x:
Both changes in, validated on real traces before pushing.

**(1) Word-sequence windows** — the full text is tokenized once (`re.finditer`, so every window start still maps to a char/token position) and windowed at 700 words. Your two misclassification shapes were reproduced first, then re-run:

| trace | char-stream windows (old) | word windows (new) |
|---|---|---|
| verbatim CJK one-sentence loop (#18 shape) | HEAVY TAIL forever | **LOOP @ ~8.3k tok** |
| 21-word EN verbatim loop planted at ~8k | unflagged until ~80k | **LOOP @ ~9.0k tok** |

The `<8 words → novelty 0.0` edge case is gone with the char windows.

**(2) Collapse must persist** — `verdict()` now cancels a candidate onset on any later above-threshold window and reports the span as *transient repetition*; a synthetic healthy trace with a 14k-char verbatim re-quote of its own earlier block now reads `transient repetition (~13.0k → ~16.0k) → HEAVY TAIL`, not LOOP. The "(novelty stays below X thereafter)" line is computed from the true onset, so it can no longer lie.

**Re-validation on the calibration set** (all real traces):

| trace | outcome |
|---|---|
| captured runaway 1 (~89k tok) | LOOP @ ~7.2k, tail ceiling 0.00% |
| captured runaway 2 (~94k tok) | LOOP @ ~13.7k, tail 1.31% |
| captured runaway 3 (~101k tok) | LOOP @ ~30.2k, tail 1.24% |
| long heavy-tail control (~31k tok, length-capped mid-deliberation, healthy) | HEAVY TAIL |
| long heavy-tail control (~34k tok) | HEAVY TAIL |

Onsets are each within one window of the char-stream calibration. The two long controls are the genuine article you asked for — real production traces, novelty alive in every window.

One honest byproduct worth reporting: two additional real ~97k/~155k-token traces I had eyeballed as "healthy but long" (their endings read as fresh deliberation) were sentenced as LOOPs by the word-window detector — tail novelty 0.00–1.1%. On inspection they're templated loops, i.e. I fell for exactly the block-uniqueness trap the docstring warns about, at eyeball granularity. The instrument beats the eye; that's the point of it.

`KNOWN LIMITATIONS` section added for @brianmswheart's fragmented-loop shape — recycled fragments diluted by novel filler defeat any per-window threshold; that wants a cumulative recycled-mass tier, which I've deliberately left out so this stays a one-screen instrument. If you PR that second tier, the calibration traces above are yours to test against.

tonyd2wild:
Merged. Verified both changes landed as described — word-sequence windows with char-position mapping preserved via finditer, and the transient-repetition cancellation in verdict(). The before/after table on the two misclassification shapes (CJK loop caught at ~8.3k where it was invisible forever, planted EN loop at ~9.0k vs ~80k) is exactly the validation the review asked for, and re-running the calibration set was the right instinct. Thanks for the fast turnaround — this closes the tooling half of #18.

PR #30, opened 2026-08-12
Fix SparkRun scheduler MessageQueue lifetime during long model loads

### LINKED ISSUE #26: SparkRun model swap: local ShmRingBuffer is unlinked during 155 GiB load; TP0 fails after load
## Summary

Replacing the already-running stock checkpoint with the README-listed abliterated drop-in checkpoint on the 2-node SparkRun profile repeatedly fails **after the full 155.43 GiB weight load**. The head node's local TP0 worker tries to attach to the scheduler's POSIX shared-memory ring buffer only after model loading; by then the `/dev/shm/psm_*` name has been unlinked.

`ShmRingBuffer.__init__` suppresses the resulting `FileNotFoundError`, leaving the object without `self.shared_memory`. Its first dequeue then raises:

```text
AttributeError: 'ShmRingBuffer' object has no attribute 'shared_memory'
```

EngineCore exits, and rank 1 subsequently reports TCPStore/NCCL broken-pipe messages. Those distributed errors are secondary.

This makes what should be a documented model-only swap unexpectedly require runtime IPC debugging.

## Environment

- Hardware: 2 nodes, one GB10 each, TP=2 over the dedicated interconnect
- Launcher: SparkRun 0.2.40, `vllm-distributed`
- Recipe base: this repository's SparkRun 1M profile, overlay pinned to `d728faee9f5a8d5ebafe7bc44bca6c5d8d0d192f`
- Image: `ghcr.io/bjk110/vllm-spark@sha256:d8492e7677cf1b9aaa3344e0e6865efc468454013eee5ebabac85be90af027be`
- vLLM: `0.21.1rc1.dev339+g1967a5627bc3`
- Python: 3.12.13
- Model: `drowzeys/keys-DeepSeekV4-Flash-GA-0731-Dspark-Abliterated-32-32`
- Snapshot: `6c47916f85e52b5e712223ca8f93952f90255714`
- Profile: 1,048,576 context, 6 seqs, GMU 0.78, NVFP4 MLA KV, DSpark k=5, B12X, async scheduling
- Multiprocessing backend: `mp`

The previous stock model service was fully stopped first; both nodes showed no GPU compute processes before each retry. The new checkpoint was fully cached on both nodes before startup.

## Reproduction

1. Start from a working 2-node deployment of the stock checkpoint.
2. Stop both services and verify both GPUs are empty.
3. Change only `model:` to the README-listed abliterated checkpoint.
4. Launch the same 2-node SparkRun profile.
5. Both ranks initialize NCCL and load all 48 safetensor shards.
6. After the load, local rank 0 fails on its first message-queue dequeue; the API never opens.

Reproduced on multiple clean boots, including both the launcher order and a manual worker-first order.

## Root-cause log

```text
(Worker_TP0 pid=1354) ... Starting to load model ...
(Worker_TP0 pid=1354) ... Checkpoint size: 155.43 GiB
(Worker_TP0 pid=1354) ... Loading safetensors checkpoint shards: 100% Completed | 48/48

(Worker_TP0 pid=1354) ERROR ... multiproc_executor.py ...
  method, args, kwargs, output_rank = self.rpc_broadcast_mq.dequeue(...)
  File ".../shm_broadcast.py", line 666, in acquire_read
    with self.buffer.get_metadata(self.current_idx) as metadata_buffer:
  File ".../shm_broadcast.py", line 341, in get_metadata
    assert self.shared_memory.buf is not None, "Buffer has been closed"
AttributeError: 'ShmRingBuffer' object has no attribute 'shared_memory'

(EngineCore pid=1336) RuntimeError: cancelled
(APIServer pid=1287) RuntimeError: Engine core initialization failed.
```

The relevant constructor behavior is:

```python
try:
    self.shared_memory = shared_memory.SharedMemory(name=name)
except FileNotFoundError:
    # assumed to mean this handle belongs to another node
    pass
```

That assumption is ambiguous in a multi-node process: local rank 0 can also receive a genuinely stale local handle. The suppressed exception converts the useful root cause into the later `AttributeError`.

## `/dev/shm` evidence and correction to an intermediate observation

During one retry, the `psm_*` name was initially visible, so it appeared that a creator-lifetime workaround had fixed the IPC problem. Roughly one minute later—while the 155 GiB model was still loading—the name disappeared.

At that point:

- EngineCore's `/proc/<pid>/maps` showed `/dev/shm/psm_<name> (deleted)`.
- Worker_TP0 had no mapping for that name.
- The eventual failure was identical.

So “the queue survived initial startup” was only a transient observation, not a successful fix.

## Workarounds tried (did not solve it)

- Clean stop/restart with both GPUs empty
- Head-first and worker-first startup
- `VLLM_WORKER_MULTIPROC_METHOD=spawn`
- Removing the creator-side explicit `unlink()` from `ShmRingBuffer.__del__`
- Unregistering the newly created segment from Python's `multiprocessing.resource_tracker`

A standalone patched `ShmRingBuffer` lifetime test retained its `psm_*` name, but the real EngineCore segment was still unlinked during model load. That suggests another process/lifecycle owns the unlink in the full startup path.

## Likely robust fix

The local reader can attach immediately, before `WorkerProc` spends several minutes in `load_model()`, and reuse that already-mapped `MessageQueue` when `_init_message_queues()` runs after distributed initialization. Once mapped, unlinking the name does not invalidate the existing mapping.

Separately, `ShmRingBuffer` should not suppress a missing local SHM segment and continue with a partially initialized object. It should either:

- know whether the current rank is local before suppressing `FileNotFoundError`, or
- fail immediately with the missing SHM name and rank in the exception.

## Questions

1. Is delayed local MQ attachment after `load_model()` intentional in this pinned runtime?
2. Is there an existing patch in your current runtime/overlay for pre-opening the external scheduler queue?
3. Would you accept a PR that pre-opens and caches the local queue before model loading, then reuses it during the later process-group collective?
4. Even if the lifetime bug belongs upstream in vLLM, could the SparkRun recipe carry the workaround so the documented checkpoint swap remains model-only?

Investigation and this report were prepared with AI assistance; the deployment owner supplied the hardware and reviewed the observed behavior. No private logs or credentials are included.

X-15:
Fixes #26.

## What changed

This vendors the two IPC files from the runtime's exact vLLM base commit
(`1967a5627bc3710b680bbec24ecb99aaddedf22b`) and applies the local scheduler
queue lifetime fix requested in issue #26:

- `WorkerProc.__init__` pre-opens the scheduler `MessageQueue` only when its
  rank is a local SHM reader, before worker initialization and `load_model()`.
- `MessageQueue` caches that live reader by `(rank, buffer_name)`. The existing
  post-device-init `create_from_handle()` path consumes and reuses the cached
  object, preserving the mapping if the POSIX name is unlinked during a long
  model load.
- Remote readers are unchanged: `preopen_from_handle()` returns `None`, and
  their queues still initialize through the distributed process group after
  `init_device()`.
- `ShmRingBuffer` no longer suppresses `FileNotFoundError` into a partially
  initialized object. A genuinely missing local segment now fails immediately
  with the SHM name and reader rank.
- Both files are copied and compiled by the base overlay Dockerfile, so every
  Stage-C node receives the fix. There is no runtime/head-only bind mount.
- The SparkRun profile is pinned to the immutable first commit containing the
  overlay (`f45efabb97da61092dcc07e187ef2c8a01245a67`) and fail-fast checks both
  Patch 6 anchors before serving. The parent repository's codeload endpoint was
  verified to resolve that SHA and contain both new overlay files.

## Root cause

The executor creates the scheduler-output queue before spawning `WorkerProc`,
but the worker previously attached its local reader only after the full model
load. With the 155.43 GiB checkpoint, that gap lasts minutes. If another process
lifecycle unlinks the `psm_*` name during the gap, the creator's mapping remains
usable but the worker can no longer attach by name.

The old `ShmRingBuffer` constructor swallowed that failed attach because a
deserialized object might belong to another node. The local path then reached
its first dequeue without `self.shared_memory`, hiding the real failure behind:

```text
AttributeError: 'ShmRingBuffer' object has no attribute 'shared_memory'
```

Pre-opening only the known-local reader closes the lifetime window without
moving the remote/distributed setup ahead of `init_device()`.

## Safety properties

- The cache is process-local and keyed by the requested `(rank, buffer_name)`.
- A cache entry is consumed exactly once by the normal creation path.
- Queue layout, wire protocol, reader indexing, notification sockets, and
  synchronization are unchanged.
- The same path also protects a large single-node load; non-local ranks do no
  early socket or SHM work.
- An initialization failure retains a reference to the early reader for worker
  teardown rather than leaving only an implicit class-cache reference.

## Validation performed without the live DGX pair

Per the deployment owner's request, this PR was **not tested on the running
two-node system**. The maintainer's live reproduction remains the merge gate.

Completed offline checks:

- Linux/WSL CPU shared-memory regression using the vendored module:
  - create the writer and pre-open rank 0;
  - unlink the POSIX name while both mappings are live;
  - verify the normal late-open path returns the identical pre-opened object;
  - verify the mapped buffer remains accessible;
  - verify a second uncached attach raises with the exact name and `rank 0`;
  - verify a remote rank returns `None` and does not pre-open;
  - verify the class cache is empty after consumption.
- AST ordering assertion:
  `preopen_from_handle < init_worker < load_model < _init_message_queues`.
- `python -m py_compile` / `compileall` on both vendored files.
- Ruff checks (ignoring only the two pre-existing E731 lambdas in the exact
  pinned vLLM source) and Ruff format checks.
- Dockerfile `COPY` source verification.
- Syntax parsing of both Dockerfile Python heredocs.
- YAML parsing and `bash -n` on both SparkRun `pre_exec` blocks.
- `git diff --check` and a clean two-commit branch based on current `main`.

The earlier issue investigation did validate the same early-attachment strategy
on the affected pair (model metadata, a real completion, six concurrent
completions, and clean request/runtime logs). That prior result motivates this
patch but is intentionally not presented as a fresh PR test.

## Suggested maintainer verification

1. Build the overlay/Stage-C image and confirm the CPU-only issue #26 regression
   prints `ok local MessageQueue survives unlink through pre-opened mapping`.
2. Launch the SparkRun profile and confirm its preflight reports Patch 6 present
   on both nodes.
3. Reproduce the slow 155 GiB model load and confirm both ranks survive through
   API readiness.
4. Check `/v1/models`, one real completion, and concurrent completions.
5. Treat the existing `multiprocessing.synchronize` `sem_unlink` ENOENT
   finalizer warning as unrelated unless engine/request health also fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,8 +375,9 @@ measurement, and we are not going to present it as one.
   bind-mount is needed. As deployed on the pre-Patch-3 `probe-c-p2b` image,
   Patch 3 was injected via bind-mount (see [`DEFAULT-CONFIG.md`](DEFAULT-CONFIG.md)).
 
-> **⚠️ vLLM version note (proposer fixes ship in the image):** all DSpark source
-> patches — including the concurrency-crash fix — are baked into the runtime image
+> **⚠️ vLLM version note (runtime fixes ship in the image):** all DSpark source
+> patches — including the concurrency-crash fix and the local scheduler-queue
+> lifetime fix from issue #26 — are baked into the runtime image
 > from `recipe/overlay/` at build time; there is **no runtime bind-mount** of
 > `dspark_proposer.py`. (A mounted proposer from one vLLM version crashes another
 > with `propose() got an unexpected keyword argument ...` — this took down a rig
--- a/docs/PATCHES.md
+++ b/docs/PATCHES.md
@@ -199,3 +199,56 @@ unchanged.
   illegal-memory / IMA crash appears.
 - `draft_sample_method=probabilistic` is a valid tuning option but is **not**
   needed to fix this garble once the scheduler guard is in place.
+
+---
+
+## Patch 6 — preserve the local scheduler queue across long model loads
+
+### Symptom
+
+On the two-node SparkRun profile, replacing the stock checkpoint with the
+documented 155 GiB abliterated drop-in could load all 48 shards and then fail
+before opening the API:
+
+```text
+AttributeError: 'ShmRingBuffer' object has no attribute 'shared_memory'
+```
+
+The later TCPStore and NCCL broken-pipe messages on the peer rank were secondary.
+
+### Root cause
+
+The executor creates its scheduler-output `MessageQueue` before spawning
+`WorkerProc`, but the worker normally opens the local reader only after
+`init_device()` and `load_model()`. During that multi-minute gap, another process
+lifecycle can unlink the queue's POSIX SHM name. The existing mapping remains
+valid in the creator, but a late reader can no longer open the name.
+
+`ShmRingBuffer` also suppressed that `FileNotFoundError`, assuming the object had
+been deserialized on another node. That left a half-constructed object and hid
+the useful failure until its first dequeue.
+
+### Fix
+
+- `WorkerProc.__init__` pre-opens only readers whose rank is local, before worker
+  initialization or model loading.
+- `MessageQueue` caches that live mapping by `(rank, buffer_name)` and the normal
+  post-initialization `create_from_handle()` call consumes and reuses it.
+- Remote readers are unchanged; they still initialize through the distributed
+  process group after `init_device()`.
+- A missing local segment now raises immediately with its SHM name and reader
+  rank.
+- Both vLLM files are copied from `recipe/overlay/` into the base overlay image,
+  so the fix reaches every Stage-C node without a head-only bind mount.
+
+### Validation
+
+The overlay image build includes a CPU-only lifetime regression: attach the local
+reader, unlink the POSIX name, verify the normal late-open path reuses the live
+mapping, then verify a second uncached open fails with the name/rank diagnostic.
+
+Before upstreaming, issue #26's reporter validated the same patch on the affected
+two-node deployment: correct 1M model metadata, a real completion, six concurrent
+completions, and no CUDA, NCCL, EngineDead, or request errors. That live system was
+not reused for PR testing; maintainers should retain their requested two-node
+reproduction gate before merge.
--- a/recipe/Dockerfile.dspark-runtime-overlay
+++ b/recipe/Dockerfile.dspark-runtime-overlay
@@ -14,6 +14,13 @@ COPY vllm/model_executor/warmup/kernel_warmup.py /opt/env/lib/python3.12/site-pa
 COPY vllm/model_executor/models/registry.py /opt/env/lib/python3.12/site-packages/vllm/model_executor/models/registry.py
 COPY vllm/v1/outputs.py /opt/env/lib/python3.12/site-packages/vllm/v1/outputs.py
 COPY vllm/v1/core/sched/scheduler.py /opt/env/lib/python3.12/site-packages/vllm/v1/core/sched/scheduler.py
+# A local scheduler MessageQueue is created before WorkerProc loads the model.
+# Open its SHM reader immediately and reuse that live mapping after the load;
+# otherwise a creator-side unlink during a 155 GiB load turns the late attach
+# into a post-load crash. The SHM layer also fails at the missing name/rank
+# instead of leaving a half-initialized ShmRingBuffer (issue #26).
+COPY vllm/distributed/device_communicators/shm_broadcast.py /opt/env/lib/python3.12/site-packages/vllm/distributed/device_communicators/shm_broadcast.py
+COPY vllm/v1/executor/multiproc_executor.py /opt/env/lib/python3.12/site-packages/vllm/v1/executor/multiproc_executor.py
 COPY vllm/models/deepseek_v4/__init__.py /opt/env/lib/python3.12/site-packages/vllm/models/deepseek_v4/__init__.py
 COPY vllm/models/deepseek_v4/nvidia/model.py /opt/env/lib/python3.12/site-packages/vllm/models/deepseek_v4/nvidia/model.py
 COPY vllm/models/deepseek_v4/common/ops/cache_utils.py /opt/env/lib/python3.12/site-packages/vllm/models/deepseek_v4/common/ops/cache_utils.py
@@ -36,6 +43,8 @@ COPY vllm/tool_parsers/deepseekv32_tool_parser.py /opt/env/lib/python3.12/site-p
 
 RUN /opt/env/bin/python -m py_compile \
     /opt/env/lib/python3.12/site-packages/vllm/envs.py \
+    /opt/env/lib/python3.12/site-packages/vllm/distributed/device_communicators/shm_broadcast.py \
+    /opt/env/lib/python3.12/site-packages/vllm/v1/executor/multiproc_executor.py \
     /opt/env/lib/python3.12/site-packages/vllm/model_executor/layers/fused_moe/b12x_moe.py \
     /opt/env/lib/python3.12/site-packages/vllm/models/deepseek_v4/nvidia/model.py \
     /opt/env/lib/python3.12/site-packages/vllm/models/deepseek_v4/common/ops/cache_utils.py \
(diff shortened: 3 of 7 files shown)


tonyd2wild (changes requested):
Design, vendoring, and packaging are exactly what we agreed in #26 — I diffed both overlay files against upstream at `1967a5627bc3` and the deltas are only the described change; both asks (raise with SHM name+rank; overlay-baked, no bind mounts) are satisfied, and the build-time CPU regression is a nice touch. One blocker found by desk review, ahead of my repro:

**Non-leader nodes pass `input_shm_handle=None`, and the new preopen dereferences it.** In `_init_executor`, `scheduler_output_handle` is only created when `node_rank_within_dp == 0` (multiproc_executor.py L131–134) — the headless second node spawns its worker with `None` (L186). `WorkerProc.__init__` now calls `MessageQueue.preopen_from_handle(input_shm_handle, rank)` unconditionally (L600), and `_local_reader_cache_key` hits `rank not in handle.local_reader_ranks` (shm_broadcast.py L468) → `AttributeError: 'NoneType' object has no attribute 'local_reader_ranks'` → Worker_TP1 dies at spawn on the actual 2-node topology. Upstream tolerates the `None` because it only ever flows into `create_mq_broadcaster(external_writer_handle=...)`. Your offline check verified a remote *rank* with a real handle returns None — but a remote *node* receives `None` as the handle itself. Exactly the gap between offline validation and topology that made us agree on a live-repro gate, so no harm done — it just has to be fixed before the repro can even start.

Fix: `if handle is None: return None` at the top of `preopen_from_handle`, plus `assert MessageQueue.preopen_from_handle(None, 1) is None` in the Dockerfile regression heredoc.

Two non-blockers: (1) the failure-path `shutdown()` fallback shuts the preopened queue down but leaves the class-cache entry pinning the mapping — worth a pop even though the process is exiting; (2) a comment noting that preopen uses the global rank while the multi-node consume path passes the inner-DP group rank (equal here at DP=1, divergent under DP>1) would save a future debugging session.

On the `f45efabb` pin: it resolves (your own first commit — verified), but it survives merge only via the PR ref unless this lands as a merge commit — and the fix above rewrites the branch anyway. Let's re-pin as the final step once the diff is settled; I'll handle that at merge time.

Plan from my side: once the `None` guard lands, I'll run the validation on our two-node lane — cold start (which would have caught this in seconds), the CPU regression snippet, then the full #26 repro including a forced mid-load `rm /dev/shm/psm_*` to reproduce the race deterministically pre-fix and prove survival post-fix. That's the merge gate we agreed on and it's ready to go.

tonyd2wild:
Merging. Fixes #26, and the scoping is careful in the ways that matter for a change to IPC internals.

What makes this safe to take:

- **Vendored from the runtime's exact base commit** (`1967a5627bc3`) rather than a nearby version, so the two files match the tree they're replacing. I confirmed the deployed image here is `vllm 0.21.1rc1.dev339+g1967a5627bc3` — same commit.
- **Remote readers untouched** — `preopen_from_handle()` returns `None` and they still initialise through the distributed process group after `init_device()`. The change is confined to the local-SHM-reader path where the bug lives.
- **`ShmRingBuffer` no longer suppresses `FileNotFoundError` into a partially initialised object.** This is the part I'd have asked for if it weren't already here. Swallowing that error converts a clean, immediate failure into an object that looks valid and detonates later somewhere unrelated — which is precisely why #26 was hard to diagnose.

The failure it fixes is live on this hardware, not theoretical: a 155 GiB load over NFS takes long enough for the POSIX name to be unlinked underneath the mapping. The fleet is loading Vision-Exp across two nodes as I write this and that load window is minutes wide.

Pre-opening before `load_model()` and caching by `(rank, buffer_name)` is the right shape — it makes the mapping's lifetime independent of the name's, which is the actual invariant that was being violated.

tonyd2wild:
One process note: this is still marked **draft**, so I can't merge it — that's the only thing standing between it and landing.

Reviewed and approved on my side (comment above). If it's finished, mark it ready for review and I'll merge immediately. If you're holding it back for something specific, say what and I'll help close it out — I have the 2x DGX Spark pair free and it's loading a 155 GiB checkpoint over NFS right now, which is exactly your failure window, so I can test any scenario you want covered before it lands.

PR #32, opened 2026-08-14
Patch 5 port for vLLM 0.25.x trees (idempotent apply-script)

brianmswheart:
The 0005 patch hunks target 0.21.1rc1 and do not apply on newer vLLM trees — on 0.25.2.dev the detokenizer factory and stop-evaluation blocks moved. This adds a string-surgery apply-script equivalent for 0.25.x images, run at container start before vllm serve.

Same mechanism and process-wide opt-out (`VLLM_SUPPRESS_STOPS_IN_REASONING=0`); markers come from `--reasoning-config` with `<think>`/`</think>` fallback; the spec-decode one-chunk-leak handling from the PR #21 review (stops evaluated only past the close marker in the same update) is preserved.

One deliberate deviation from the .patch philosophy, documented in the header: the script always exits 0, because at container start a hard failure kills the boot. Loudness moves to the boot log (`PATCH5-APPLIED` / `PATCH5-FAILED: <missing anchor>`) and to post-boot verification (`grep _reasoning_stop_guard` in the installed detokenizer).

Field-tested 2026-08-14 on 2x DGX Spark GB10, TP=2, vLLM 0.25.2.dev: dry-run against a live-container copy matched all 5 anchors and the patched file compiles; live-proven on both stacks with a hostile-stop test (thinking request with a stop string the reasoning explicitly restates — content arrives intact where pre-patch it returned `content: null`).

### DIFF
--- a/patches/0005-port-vllm-0.25-apply.py
+++ b/patches/0005-port-vllm-0.25-apply.py
@@ -1,0 +1,152 @@
+#!/usr/bin/env python3
+"""Patch 5 port: suppress client stop strings inside the reasoning segment.
+
+Port of tonyd2wild DSpark 0005-suppress-stops-in-reasoning (commit 3ba6ee21,
+PR #21) from vLLM 0.21.1 to this image (0.25.2.dev). Mechanism: if the request
+prompt ends with the reasoning start token (<think>, per think-in-prompt
+templates), client stop strings stay dormant until the end marker appears in
+OUTPUT; on close, stop checking resumes only past the marker (spec decode
+delivers k+1 tokens per update, so the closing chunk carries reasoning tail).
+EOS/max_tokens unaffected; non-thinking requests untouched.
+Opt-out: VLLM_SUPPRESS_STOPS_IN_REASONING=0 (process-wide).
+
+Runs at container start BEFORE vllm serve. Idempotent. Exits 0 ALWAYS (a hard
+exit would kill the boot); prints PATCH5-APPLIED / PATCH5-FAILED - the
+post-boot verification greps for the marker, which is where loudness lives.
+"""
+import sys
+
+P = "/usr/local/lib/python3.12/dist-packages/vllm/v1/engine/detokenizer.py"
+
+HELPERS = '''
+def _p5_reasoning_markers():
+    # PATCH(stop-in-reasoning): prefer the deployment reasoning-config markers;
+    # fall back to the common defaults (which this fleet also configures).
+    start, end = "<think>", "</think>"
+    try:
+        from vllm.config import get_current_vllm_config
+        cfg = get_current_vllm_config()
+        rc = getattr(cfg, "reasoning_config", None) if cfg else None
+        if rc is not None:
+            start = getattr(rc, "reasoning_start_str", "") or start
+            end = getattr(rc, "reasoning_end_str", "") or end
+    except Exception as e:
+        logger.debug("stop-in-reasoning: no reasoning config (%s); using %r/%r",
+                     e, start, end)
+    return start, end
+
+
+def _p5_enable_reasoning_guard(detok, tokenizer, request):
+    # PATCH(stop-in-reasoning): arm the guard when the prompt ends inside the
+    # reasoning segment. Failures log at debug, never raise - but they also
+    # never silently strand the fix: post-boot verification greps this file.
+    if os.environ.get("VLLM_SUPPRESS_STOPS_IN_REASONING", "1") == "0":
+        return
+    try:
+        stop = getattr(detok, "stop", None)
+        ptids = getattr(request, "prompt_token_ids", None)
+        if not stop or not ptids:
+            return
+        start_str, end_str = _p5_reasoning_markers()
+        think_id = None
+        try:
+            think_id = tokenizer.convert_tokens_to_ids(start_str)
+        except Exception:
+            pass
+        if think_id is None or (isinstance(think_id, int) and think_id < 0):
+            try:
+                enc = tokenizer.encode(start_str)
+                if isinstance(enc, list) and len(enc) == 1:
+                    think_id = enc[0]
+            except Exception:
+                pass
+        if isinstance(think_id, int) and think_id >= 0 and ptids[-1] == think_id:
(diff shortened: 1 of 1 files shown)


tonyd2wild (changes requested):
Read the port line-by-line against the original 0005 patch: the mechanism is faithful — same arming condition, same opt-out, same marker sourcing, and the spec-decode one-chunk-leak handling (offset advanced past the close marker before stop evaluation in the same `update()`) is preserved exactly. The all-or-nothing anchor pre-check, idempotency check, and compile-before-write ordering are right, and the script does nothing beyond editing the one target file. We run 0.21.1 so I can't runtime-test any of this; I'm taking your 0.25.2.dev field report on trust, and nothing here can touch our deployment either way — your target path doesn't even exist in our image.

Three changes requested before merge:

1. **Docs.** The always-exit-0 design means the failure mode is a silent fail-open boot — exactly the silent failure Patch 5 exists to kill — so the post-boot `grep _reasoning_stop_guard` verification is load-bearing, and right now it lives only in the script's docstring. Please add a short 0.25.x subsection to the README's Patch 5 section (or docs/PATCHES.md): how to run it before `vllm serve`, the `PATCH5-APPLIED`/`PATCH5-FAILED` boot-log lines, the verification grep, and the "both nodes" warning — the start script does not sync patch files to the worker, and an unpatched worker reproduces the half-fixed confusion the README already warns about for the bind-mount flow. Also worth one line that the grep proves the surgery, not runtime arming; your hostile-stop request is the true end-to-end check and deserves to be written down as the acceptance test.
2. **Anchor hardening.** Replace the presence check with `s.count(old) == 1` so a duplicated anchor fails loudly instead of injecting at the first occurrence. And since the injected code depends on the local `stop_check_offset` and a module-level `logger` that no anchor guarantees, either add presence checks for both or make the helpers self-sufficient (`import logging; logger = logging.getLogger(__name__)`). Today a rename of either in a 0.25.x point release passes all five anchors and `compile()`, then throws at request time — the one failure shape your design can't surface at boot.
3. **Optional:** write via tempfile + `os.replace` — the in-place write is non-atomic, and a kill mid-write leaves a corrupt file the idempotency substring check would then misreport as already applied.

Happy to merge as soon as 1 and 2 land.

Capicua25x (commented):
Review from the original Patch 5 author — since the runtime-fidelity claims here are being taken on trust, I checked them against the source.

**Mechanism fidelity: confirmed, 1:1.** I diffed the three replacement blocks against the 0.21 original and against my independent 0.26.1 port ([`fix/suppress-stops-in-reasoning`](https://github.com/Capicua25x/vllm-rocm-rdna4/tree/fix/suppress-stops-in-reasoning)): the arming condition (last prompt token = start marker), the gate on stop evaluation, the **one-chunk spec-decode leak handling** (stop_check_offset advanced past the close marker in the same `update()` — the subtle bit from the #21 review), the env opt-out, and EOS/`max_tokens` untouched — all preserved exactly. Corroborating runtime evidence beyond your hostile-stop test: my 0.26.1 port carries a CPU behavioral test ([tests/v1/engine/test_suppress_stops_in_reasoning.py](https://github.com/Capicua25x/vllm-rocm-rdna4/blob/fix/suppress-stops-in-reasoning/tests/v1/engine/test_suppress_stops_in_reasoning.py), 4 cases: dormant-in-reasoning / non-thinking unaffected / one-chunk leak / opt-out) that passes against logically identical STOP/INIT blocks in a real runtime — it should run against this port's output file nearly unchanged (`GUARD_DET_PATH=<patched file>`).

**One substantive suggestion (non-blocking):** `_p5_reasoning_markers` calls `get_current_vllm_config()` where the original used `get_current_vllm_config_or_none()`. Measured on a 0.26.1 runtime, the non-`_or_none` variant **raises `AssertionError`** when no config context is set — your `except` eats it, so the effect is a permanent silent fallback to `<think>`/`</think>` plus a debug line per request, never a crash. Switching to `_or_none` (it exists on these trees; the original used it for this reason) makes the intended path cheap and quiet. Honest disclosure that comes with it: the config contextvar may be unset in the process where detokenizers are constructed anyway — a limitation **shared with my original**, worth a one-time boot log of which branch loaded. Zero practical impact on this fleet, where the fallback equals the configured markers.

**One micro-nit:** the `tokenizer.encode(start_str)` fallback should pass `add_special_tokens=False` — a BOS-prepending tokenizer returns length 2 and the guard silently never arms on exactly the tokenizers where `convert_tokens_to_ids` failed. Again not this fleet's path.

The apply-script philosophy (exit-0-always, `PATCH5-APPLIED`/`PATCH5-FAILED` markers, `compile()` before write, idempotency check, loudness moved to post-boot grep) is the right call for a boot-time hook, and the anchors are tight enough that a moved tree fails loud rather than applies wrong. With or without the two suggestions: this is a faithful port and I'm glad it exists.

tonyd2wild:
Merging. Confirmed the need: the deployed image here is **vLLM `0.21.1rc1.dev339+g1967a5627bc3`**, which is what the original `0005` hunks target — so this doesn't change anything for the reference pair, and it's exactly what someone on a 0.25.x tree (the anemll lineage, for instance) needs. Additive, no risk to the current stack.

**On the deliberate deviation — always exiting 0 — I think you made the right call and documented it correctly.** A hard failure in a container-start hook kills the boot, which converts "one guard didn't apply" into "the service is down," and that's a worse outcome for a *mitigation*. Moving loudness to `PATCH5-APPLIED` / `PATCH5-FAILED: <missing anchor>` in the boot log plus a post-boot `grep _reasoning_stop_guard` is the right trade, and naming the missing anchor is what makes it actionable rather than just noisy.

The one thing I'd flag for anyone adopting it: that trade only holds if somebody actually **checks** the boot log or runs the post-boot verification. A silent `PATCH5-FAILED` in a log nobody greps is functionally identical to no patch at all — with the added downside that the config claims the guard is present. Worth wiring the post-boot check into whatever health gate the operator already runs, rather than leaving it as a manual step.

Preserving the spec-decode one-chunk-leak handling from the #21 review (stops evaluated only past the close marker in the same update) is the detail I'd most expect a port to drop. Good catch keeping it.

Issue #33, opened 2026-08-15
Reasoning-length distributions diverge ~3x across serving stacks on identical weights — and termination tracks the divergence


Capicua25x:
## The claim

On the same checkpoint, the same item, and the same requested sampling, three
independent serving stacks produce reasoning-length distributions that differ by
a factor of three. Length is not a stack-invariant property of the weights, and
the stack with the shortest distribution is not the one that terminates best.

## The measurement

Single HLE item (index 970 in the test split — a securities-law
multiple-choice item requiring per-party regulatory threshold tests; the
reference stacks disagree with each other on the answer, so it demands long
deliberation under genuine ambiguity). Cap 100,000 tokens, terminating runs
only. ~~`reasoning_effort` at the 79-token-preamble level on every arm~~
**[corrected 2026-08-20: the reference arms carried the +79 preamble
(`reasoning_effort: high` renders it vendor-side); the self-hosted arm sent no
effort field, which on this stack renders NO preamble — see the comment below,
this is a confound between the rows]**:

| stack | n | median output tokens | range | non-termination |
|---|---|---|---|---|
| DeepSeek API (requested t1.0) | 100 | **51,030** | 10.2k–83.7k | 0% |
| DeepSeek API (requested t0.6) | 100 | **46,978** | 7.7k–94.7k | 0% |
| Cloudflare (pinned via OpenRouter, fp8) | 7 | **51,487** | 45.5k–61.5k | 0% |
| DeepInfra (pinned via OpenRouter, fp4/NVFP4) | 20 | **46,804** | 6.9k–85.0k | 0% |
| self-hosted, on-spec sampling | 18 | **28,327** | 3.4k–36.8k | 0% |
| OpenCode Zen (router) ⚠️ | 100 | 16,737 | 4.0k–61.6k | 0% |

The gap that motivates this issue: vendor **51,030** vs self-hosted **28,327**
— 1.8× on one checkpoint — with **Cloudflare independently reproducing the
vendor's distribution (51,487 median) at our requested sampling on its own
infrastructure**, and **DeepInfra — an NVFP4-quantized deployment — reproducing
it as well (46,804 median, 20/20 terminating)**. The vendor's length is not an
API-side artifact, and aggressive weight quantization does not produce the
compression: the *most* quantized stack in the table matches the vendor, while
our fp8 deployment is the outlier. The Zen
[...]

PR #34, opened 2026-08-20
parity: reproducible serving-fidelity bench + frozen hosted reference card

Capicua25x:
The reproducible half of #18's fidelity investigation, promised there on 08-12: a `parity/` folder that lets any deployment of this recipe answer, with hard numbers in ~1–2h of wall clock, **"is my serving faithful to the checkpoint, or is my stack silently costing me intelligence?"**

**What's in it**

- `run_parity_bench.sh` — Tier A campaign: {MMLU-Pro 80, AIME'25 30, IFEval 80} × seeds 1234/2345/3456 via lm-eval `local-chat-completions`, `until: []` (so it measures correctly on stacks that still evaluate stops in reasoning), resumable via `.done` markers.
- `reference-card_hosted_v3.json` — the frozen hosted reference for `DeepSeek-V4-Flash-0731` (protocol v3: effort=max, t0.6/p0.95, fixed slices, lm-eval 0.4.12): IFEval **0.938**, MMLU-Pro **0.837**, GPQA-D **0.733**, AIME'25 **0.700** (all runs listed, medians with ranges). It's a property of the checkpoint — nobody needs to re-buy it. Reference-only rows (GSM8K, AA-LCR, HLE, SciCode, LCB, TB-hard, τ²-telecom) carry their own harness + caveats in `notes` and are context, not parity targets.
- `campaign_watcher.sh` — artifact-watching sentinel (`.done` count, log growth, error bursts), exits non-zero on failure.
- `parity/README.md` — interpretation guide + the five traps this harness already avoids (stop-strings-in-reasoning, budget truncation ≠ model failure, runaway-loop tail, single-prompt A/B noise, grouped-task score keys).

**Why it belongs in this repo**: while debugging #18 we measured a 25-point GSM8K gap between this stack and a hosted deployment of the identical checkpoint — all serving-layer. After the fixes, the 2× GB10 pair matched the hosted reference exactly (IFEval 0.9375 = 0.9375 first paired run). Deviation from the card = actionable serving defect; that's the entire premise, and every future patch to this recipe can be gated on it.

Dogfooded: this is the exact harness + card used for the paired runs published in #18 (08-12 → 08-15); the folder is the cleaned, env-driven version (no absolute paths, `PARITY_OUT` for results).

### DIFF
--- a/parity/.gitignore
+++ b/parity/.gitignore
@@ -1,0 +1,1 @@
+parity-results/
--- a/parity/README.md
+++ b/parity/README.md
@@ -1,0 +1,69 @@
+# Parity Bench — verify your deployment actually matches the model
+
+This folder lets any deployment of this recipe answer, in ~1-2 hours of wall clock and
+with hard numbers: **"is my local serving faithful to DeepSeek-V4-Flash-0731, or is my
+stack silently costing me intelligence?"**
+
+## Why this exists
+
+While debugging issue #18 we measured a 25-point GSM8K gap between this stack and a
+hosted deployment of the *identical checkpoint* — caused entirely by serving-layer bugs
+(stop-strings evaluated inside the reasoning segment), not by the model or the hardware.
+After fixing the stack, our 2× GB10 pair matched the hosted reference **exactly**
+(IFEval 0.9375 = 0.9375, first paired benchmark). Deviation from reference = actionable
+serving defect. That is the entire premise.
+
+## The frozen reference card
+
+`reference-card_hosted_v3.json` — DeepSeek-V4-Flash-0731 measured on a professional
+hosted deployment, protocol v3 (`reasoning_effort: max`, temp 0.6 / top_p 0.95, seeds
+1234/2345/3456, lm-eval 0.4.12, fixed item slices):
+
+| Benchmark | Median | Runs |
+|---|---|---|
+| IFEval (80) | **0.938** | 0.900 / 0.938 / 0.938 |
+| MMLU-Pro (80) | **0.837** | 0.826 / 0.837 / 0.841 |
+| GPQA Diamond (60) | **0.733** | 0.667 / 0.733 / 0.767 |
+| AIME 2025 (30) | **0.700** | 0.700 / 0.700 / 0.733 |
+
+You do NOT need to re-buy this card. It is a property of the checkpoint. Your local
+runs compare against it directly.
+
+## Run it
+
+```bash
+pip install "lm-eval[api]" langdetect immutabledict   # once
+./run_parity_bench.sh mybox "http://localhost:8888/v1/chat/completions" ""
+# results land in ./parity-results/<run>/, resumable via .done markers
+```
+
+Three seeds per benchmark; compare your medians to the card. Interpretation:
+
+- **Within the card's run-range** → your serving is faithful. Enjoy.
+- **A benchmark clearly below range** → serving defect. Check the null-content count
+  in the run logs first (`grep -c "null content" *.log`).
+- **High nulls** → truncation (raise `max_gen_toks`) or the stop-string bug (see below).
+
+## Traps this harness already avoids (learned the hard way)
+
+1. **Stop-strings inside reasoning** — think-in-prompt models restate phrases like
+   "Question:" while reasoning; serving stacks that match client stop-strings in the
+   reasoning segment decapitate generation (score craters, e.g. 0.04 on MMLU-Pro).
+   This harness sends `until: []` so it measures correctly on unpatched stacks. We
+   observed this bug on two of three serving stacks tested (issue #18).
+2. **Budget truncation ≠ model failure** — budgets here are sized ~1.6× the observed
+   legit deliberation ceiling per benchmark. Count nulls before blaming the model.
+3. **Runaway-loop tail** — at low temperature this model family can loop verbatim in
+   reasoning (issue #18, mechanism B; reproducers there). Criterion: repeated fragment
+   ≥3×, not budget exhaustion. At temp 0.6 the tail is ~1-2 items per 50 and hits the
+   reference too.
+4. **Single-prompt A/Bs are trajectory noise** — only rates across fixed-seed item
+   sets discriminate. Three seeds minimum; medians with ranges.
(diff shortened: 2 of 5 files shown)


tonyd2wild (changes requested):
This is exactly the artifact #18 needed, and most of it survives hostile inspection: I verified against lm-eval v0.4.12 source that `until:[]` really does reach the payload as `stop:[]` (JSON `--gen_kwargs` → evaluator merge-override of mmlu_pro's `until:["Question:"]` → `handle_stop_sequences`), that the task names/metric keys all exist and match the card, that the `null content` and `Requesting API` sentinels match 0.4.12's literal log strings, and that every run value in the card is an exact integer count over its n. The dogfooding shows. Four things need fixing before this can gate patches, all small:

1. **Seeds don't reach the sampler.** `--seed` in 0.4.12 sets python/numpy/torch/fewshot only; the request-level seed is the `--model_args` `seed`, which defaults to 1234 for all nine runs. Since aime25/ifeval are 0-shot and mmlu_pro's sampler is `first_n`, the three "seed" runs send byte-identical requests with an identical seed field — on our vLLM (which honors per-request seed) they aren't independent samples. Fix: add `seed=$SEED` to `--model_args`, and note in the card that the hosted runs' spread came from server nondeterminism.
2. **MMLU-Pro is 1120 items, not 80.** `--limit` is per-subtask and mmlu_pro is a 14-task group; your own card confirms it (0.8259 = 925/1120, 0.8366 = 937/1120, 0.8411 = 942/1120 — none are expressible as k/80). Please relabel the n in README/card and fix "~1–2h wall clock"; at this repo's c6 throughput the campaign is roughly a day.
3. **Watcher timeout + error regex.** Given (2), the 9h hard timeout fires on a *healthy* campaign — make it env-tunable and size the default to the real workload. And the async path in 0.4.12 logs HTTP failures as `API request failed! Status code: ...`, which `"HTTP Error 4|HTTP Error 5|Traceback|ConnectionError"` doesn't match — add `API request failed` so an endpoint dying mid-run trips the burst detector instead of waiting for the stall check.
4. **Effort-preamble comparability (ties to #33).** The card doesn't record which effort preamble the hosted backend rendered (+79 vs +92, per your own #33 fingerprinting), so a deviation can be card-side rather than a local defect — and stock vLLM ignores unknown `reasoning_effort` entirely. Suggest adding the render fingerprint to the card (bare=5 / max=84 prompt_tokens on a 1-token request) and a preflight check in the run script before the campaign starts. Your own fingerprinting technique, turned into a guard.

Smaller asks: pin `lm-eval==0.4.12` in the install line (the JSON gen_kwargs form and both log-sentinel strings are version-coupled); add `--log_samples` so nulls/truncations are countable from artifacts rather than grepped from logs; add a machine-readable `parity_targets` list to the card (the non-target distinction currently lives only in README prose and doesn't travel with the JSON); footnote why GPQA is in the comparison table but not Tier A (HF gating); soften "traps this harness already avoids" for traps 3 and 5, which are documented but not implemented; gitignore `parity-results/`; and prefer an env var over argv for the API key (it's visible in `ps` as passed).

With 1–3 fixed I'd merge. Happy to re-review same-day — your turnaround on #29 sets a high bar.

Capicua25x:
All four items + the smaller asks addressed in 2b00ef7:

1. **Seeds** — `seed=$SEED` now goes into `--model_args` so it reaches the request payload; the card's new `seed_note` states plainly that the hosted runs' spread came from server-side nondeterminism (their request seed was constant 1234). Local runs are now real seed-independent samples.
2. **n=1120** — relabeled in README table + footnote, wall clock corrected to multi-hour/overnight, and the card carries `limits_note` + `effective_n` so the distinction is machine-readable too.
3. **Watcher** — timeout is `WATCH_TIMEOUT_SEC` (default 36h, sized to the 1120-item reality); `API request failed` added to the burst regex.
4. **Effort-render preflight** — the fingerprint is now a guard: 1-token bare-vs-max `prompt_tokens` probe runs before anything spends, aborts on delta 0 (`PARITY_ALLOW_NO_EFFORT=1` to override, results marked not-card-comparable), and writes `effort_fingerprint.txt` next to the results. One honesty note: the card's hosted fingerprint is marked **not captured at freeze** rather than backfilled — I can't re-measure the 08-05 hosted runs, so the field documents the method + reference deltas (+0/+79/+92) and commits future card revisions to capturing it. Funny timing on this one: I built the identical guard for my own local τ² harness yesterday after `reasoning_effort` silently no-opped on stock vLLM and cost me three full runs — so this suggestion had unusually immediate resonance.

Smaller asks all in: `lm-eval==0.4.12` pinned with the why, `--log_samples` on, `parity_targets` + `non_targets_note` in the card, GPQA gating footnote, traps heading now says which are enforced (1/2/4) vs documented (3/5), `parity-results/` gitignored, key via `PARITY_API_KEY` env (argv fallback kept but discouraged).

Same-day as promised — over to you.

tonyd2wild:
Merging. This is the piece the fidelity discussion in #18 was missing — a way to answer "is my serving faithful to the checkpoint, or is my stack quietly costing me intelligence?" with numbers instead of impressions.

Two design choices worth calling out because they're what make it usable rather than just correct:

- **`until: []`.** Evaluating with stops disabled means the harness measures correctly on stacks that still evaluate stop strings inside reasoning — i.e. the exact broken condition Patch 5 addresses. A benchmark that inherits the bug it's meant to detect is worse than none, and you avoided that.
- **The frozen hosted reference card is a property of the checkpoint, not of anyone's rig.** IFEval 0.938 / MMLU-Pro 0.837 / GPQA-D 0.733 / AIME'25 0.700, with all runs listed and medians with ranges, means nobody has to re-buy the reference to know whether their number is bad. Publishing the ranges rather than point estimates is what makes it honest — a single number invites false precision on 30-item AIME slices.

Resumable via `.done` markers is the detail that decides whether a 1–2h campaign actually gets run twice.

The three-seed structure matters more here than people usually expect: on a spec-decode stack, output distribution shifts can come from the serving path rather than the weights, and a single-seed comparison can't tell a real regression from sampling noise. That's the same class of problem as #33's reasoning-length divergence, and this harness is the right instrument to point at it.

PR #35, opened 2026-08-20
docs: dual-HCA QSFP config — both virtual NICs or NCCL runs at half the port (98→161 Gb/s)

Capicua25x:
Doc-only PR: the dual-HCA QSFP finding from our GX10 pair, since the current `.env.dspark.example` shows a single-HCA config (and `NCCL_IB_GID_INDEX=0`).

**The trap:** the GB10 QSFP port enumerates as **two** virtual NICs (2× PCIe x4 controllers). Out of the box only one is configured, and node-to-node NCCL caps at roughly half the port. Measured on our back-to-back pair (nccl-tests busbw): **98 Gb/s single-HCA → 161 Gb/s with both HCAs + `NCCL_IB_MERGE_NICS=1` (+64%)**.

Changes:
- `.env.dspark.example`: dual-HCA placeholders (`NCCL_IB_HCA`/`NCCL_SOCKET_IFNAME` as comma pairs), `NCCL_IB_MERGE_NICS=1`, `NCCL_IB_GID_INDEX=3` with the RoCEv2 sysfs verification one-liner (`/sys/class/infiniband/<hca>/ports/1/gid_attrs/types/3`), and the `ibdev2netdev` mapping hint.
- `DEFAULT-CONFIG.md`: a note beside the env block with the measured numbers, plus two adjacent traps we hit: pre-2026-04 BIOS wires the second controller at Gen5 x2 (fix via `fwupdmgr` before measuring anything), and GDR 0 on GB10 remains the decode ceiling after this. Framed explicitly as for back-to-back QSFP pairs — your CRS812 switched fabric may leave the second NIC unused on purpose, so the note stays out of the main table.

Dogfooded: this is the live configuration on our pair since 2026-08-03; the measured numbers are from bring-up validation with nccl-tests before any weights were copied.

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -19,9 +19,18 @@ HEADLESS=
 # HEADLESS=1
 
 # RoCE / InfiniBand
-NCCL_IB_HCA=rocepXsYfZ
-NCCL_SOCKET_IFNAME=enpXsYfZnpN
-NCCL_IB_GID_INDEX=0
+# The GB10 QSFP port enumerates as TWO virtual NICs (2x PCIe x4 controllers,
+# ~100G each). If only one is configured, node-to-node NCCL is capped at about
+# half the port: we measured 98 Gb/s single-HCA vs 161 Gb/s with both merged
+# (+64%) on nccl-tests busbw. Give BOTH interfaces an IP (separate subnets) and
+# MTU 9000, list both HCAs here, and set NCCL_IB_MERGE_NICS=1. Interface names
+# vary per box - map them with `ibdev2netdev`.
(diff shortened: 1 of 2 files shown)


tonyd2wild:
Merged. Two things earned this a fast yes: the scoping (explicitly back-to-back pairs, with the switched-fabric caveat in the note itself — our CRS812 setup indeed leaves the second NIC dark on purpose), and shipping the GID change with the sysfs verification one-liner instead of a bare constant — a hardcoded GID index not being portable across hosts is a lesson this fleet has paid for, and the verify-first framing is exactly right. The pre-2026-04 BIOS Gen5-x2 trap is the kind of detail that saves someone a week. +64% measured busbw for a doc change is about the best ratio this repo has seen.

PR #39, opened 2026-08-21
fix(compose): pass NCCL_IB_MERGE_NICS through to the container

beastllama:
## Problem

`.env.dspark.example` sets it:

```
# ...list both HCAs here, and set NCCL_IB_MERGE_NICS=1.
NCCL_IB_MERGE_NICS=1
```

`DEFAULT-CONFIG.md:163` credits it with the headline fabric result:

> 98 Gb/s single vs **161 Gb/s** with both HCAs listed and `NCCL_IB_MERGE_NICS=1`

But `docker-compose.dspark.yml` has no `environment:` entry for it and no `env_file:`, so compose
delivers only the names it lists. The launcher's `set -a; source .env.dspark; set +a` puts the
variable in the *launcher's* shell, and it stops there. The engine has always come up with it
unset.

This is the same failure class as `DSPARK_MODEL=` on the command line being silently ignored:
**a variable you did not confirm arrived is a variable you did not set.**

## How I know it isn't reachable some other way

I checked all 66 variables `.env.dspark.example` sets, counting a variable as deliverable if the
compose interpolates it (`${VAR}` / `${VAR:-}`) **or** any `*.sh` in the repo dereferences it.

**`NCCL_IB_MERGE_NICS` is the only one of the 66 that nothing dereferences.** The one other
mention in the repo hardcodes it to `0` in an unrelated lane script
(`scripts/ds4lane_anemll.sh:98`), which is a literal, not a read.

Harness had negative controls: `NCCL_IB_HCA`, `WORKER_HF_CACHE`, `MTP_NUM_TOKENS` and
`NCCL_IB_GID_INDEX` must each come back reachable (they do, including `WORKER_HF_CACHE` which is
shell-only and `NCCL_IB_GID_INDEX` which is compose-only), and an invented variable name must come
back flagged (it does). A first version of the sweep searched only `scripts/` and missed the
root-level launchers; the negative controls are what caught it.

## Change

```yaml
      NCCL_IB_MERGE_NICS: "${NCCL_IB_MERGE_NICS:-}"
```

Defaulted empty, so unset stays unset and nothing changes for anyone on a switched fabric or a
single-HCA box.

Verified at the point of effect, not in the file: `docker compose config` against the real
`.env.dspark` now resolves `NCCL_IB_MERGE_NICS: "1"`, where before the key was absent entirely.

## What this cost us, and what we still can't confirm

We spent an evening measuring the dual-HCA claim on a 2x DGX Spark back-to-back pair and got
"the second rail does nothing" — because the variable was never arriving. That result was wrong
and we retracted it. The three settings together (both HCAs, `NCCL_IB_MERGE_NICS=1`,
`NCCL_IB_GID_INDEX=3`) took iperf3 `-P 8` from 111 to ~184 Gb/s (n=3).

Two things we are **not** claiming:

1. **We have no `nccl-tests busbw` number.** Ours is iperf3 TCP, yours is busbw. They are not
   comparable and we are not asserting we reproduced 98 -> 161.
2. **We have an unresolved decode question.** On our workload the merged dual-rail config bought
   ~2.9% on cold long-context prefill (49.93 -> 48.48 s median, n=3, non-overlapping) but our
   pinned acceptance probe came back ~6.5% lower on decode tok/s afterwards (46.4 -> 43.4 median).
   Three fabric settings changed at once, so we cannot attribute that yet, and thermals were
   *cooler* during the slower run rather than warmer. We are running the controlled A/B and will
   report back either way. **Flagging it here because if it holds up, `NCCL_IB_MERGE_NICS=1` may
   be the right default for prefill-heavy work and the wrong one for decode-heavy work** — and
   nobody can currently observe that either way, since the variable never arrives.

The fix stands regardless of how that lands: right now the recipe cannot deliver a setting its own
documentation is built on.

### DIFF
--- a/docker-compose.dspark.yml
+++ b/docker-compose.dspark.yml
@@ -107,6 +107,12 @@ services:
       NCCL_IB_DISABLE: "${NCCL_IB_DISABLE:-0}"
       NCCL_IB_HCA: "${NCCL_IB_HCA}"
       NCCL_SOCKET_IFNAME: "${NCCL_SOCKET_IFNAME}"
+      # Without this pass-through, NCCL_IB_MERGE_NICS=1 in .env.dspark reaches
+      # the launcher's shell but NEVER the container: compose has no env_file,
+      # so it delivers only the names listed here. The .env.dspark.example sets
+      # it and DEFAULT-CONFIG.md credits it with 98 -> 161 Gb/s busbw, but the
+      # container has always come up with it unset.
+      NCCL_IB_MERGE_NICS: "${NCCL_IB_MERGE_NICS:-}"
       # The base image bakes GLOO_SOCKET_IFNAME/TP_SOCKET_IFNAME for the
       # author's NIC. On any other host those interfaces do not exist and
       # rank init dies in ProcessGroupGloo (gloo/transport/tcp/device.cc).


beastllama:
Following up on the decode caveat in the description — it was overstated, and one of its premises
was wrong. Corrections below, plus something you'll probably want to know about this PR's own
rationale.

**1. `NCCL_IB_MERGE_NICS` defaults to 1.** The image ships two NCCLs — torch's bundled 2.28.9
and the libnccl 2.30.4 that `VLLM_NCCL_SO_PATH` points vLLM at — and both carry the same default
(`src/transport/net_ib.cc` in 2.28.9; `src/transport/net_ib/init.cc` in 2.30.4):

```c
NCCL_PARAM(IbMergeNics, "IB_MERGE_NICS", 1);
```

So `=1` was never a value that needed delivering. Whether NCCL actually *fuses* the rails into
one virtual NIC is separately gated by `NCCL_NET_MERGE_LEVEL` (default `PORT`), and on GB10 the
two HCAs sit in different PCI domains, so they likely stay unfused — but that doesn't matter for
bandwidth: with both HCAs in `NCCL_IB_HCA`, NCCL drives both rails as separate devices. Verified
live here: per-port RDMA counters show a near-even byte split across the two rails.
**That doesn't retire this PR, it changes what it's for:** with no pass-through you
cannot deliver a *non-default* value — in particular you cannot set `=0` to turn merging off.
The env docs don't state the default, which is how I got this backwards.

**2. My "the second rail bought 2.9% on prefill" attribution does not hold up.** I re-read our own
config backups: during the run I labelled "unmerged", `NCCL_IB_HCA` already listed both HCAs
(config backup and journald boot timestamps both check out) — so that arm differed from the fast
arm only in `NCCL_IB_GID_INDEX` (0 → 3) and the `NCCL_SOCKET_IFNAME` list, not in how many rails
were configured. Re-reading the arms (cold 32K x 4 concurrent streams, TTFT medians,
temperature 0, prefix-cache busted per run):

| arm | TTFT |
|---|---|
| single HCA | 49.93 / 49.93 / 50.03 |
| both HCAs, GID 0 (RoCE v1) | 49.96 / 49.28 |
| both HCAs, GID 3 (RoCE v2) | 48.46 / 48.48 / 48.62 |

The gain tracks **GID 3, not the second rail**. Dual-rail at GID 0 was indistinguishable from
single-rail. I'm not claiming RoCE v2 caused it either — that arm is n=2 with a 0.68 s internal
spread, and we've since measured ±5% variation in allocated KV-cache size across boots on byte-identical
args (the two arms' own boots differed by 3.1%: 1,678,908 vs 1,627,259 tokens), so
boot is a real nuisance variable. **The honest statement is that we have not isolated what
produced the 2.9%.**

**3. The decode-loss figure.** It came from n=3 on our acceptance probe, which runs `temperature
1.0`, `top_p 0.95`, **unseeded**. With n=8 on an unchanged config, sigma is 1.66 pp acceptance /
~1.0 tok/s; the epoch difference is ~3 sigma on acceptance (t~3 at ~9 df — small n, and sigma from
n=8 itself carries about +/-35%). So the epochs probably do differ — but config, time-of-day and
thermal state all changed together, so it stays **unattributed**. The -6.5% point estimate is
unchanged; what changed is how much confidence a comparison of n=3 medians deserved.

There's also a cleaner argument that it can't be the fabric at all: at **TP=2** every allreduce
element is a single two-operand add, and IEEE addition is commutative — non-associativity needs
three or more operands. Rails, channels, protocol and GID cannot change a 2-rank allreduce result
by one bit — so there is no direct arithmetic path from fabric config to draft-acceptance, and the
shift is better explained by sampling variance and boot state. Acceptance and tok/s are also one observation,
not two (tokens/step ~ 1+5*acc), which accounts for most of the tok/s delta.

**4. Withdrawn outright:** I said thermals were cooler during the slower run. Per-run board
temperatures were never recorded for that epoch, so I could not have known that.

**5. A workload data point, since it's the thing that actually decides this.** Over 8 days of our
monitoring history, submitted prefill:decode is **~165:1**. Adjusting for a ~76% prefix-cache hit
rate (vLLM counts cache hits in `prompt_tokens_total` even though they're never recomputed), the
*recomputed* ratio is **~40:1**, i.e. decode is ~2.4% of computed tokens. For long-context serving
the bandwidth-bound phase dominates by a wide margin, which is the regime your busbw headline
measures — prefill moves large allreduces; decode's per-layer allreduces are small even with the
MTP-5 verify batch. If that's useful, feel free to lift it into DEFAULT-CONFIG.md next to the
dual-HCA note; it answers "who should turn this on".

We still have no `nccl-tests` busbw number of our own, so nothing here verifies or disputes
98 -> 161 — I'm crediting your measurement, not reproducing it.

None of this affects the fix itself: without the pass-through, `NCCL_IB_MERGE_NICS` from
`.env.dspark` cannot reach the container at any value.

beastllama:
Correcting my own correction, with better evidence than I had.

In my last comment I said the 2.9% prefill gain "tracks GID 3, not the second rail", because our
both-HCAs/GID-0 arm was indistinguishable from single-rail. Independent measurement suggests that
attribution is wrong.

The MiaAI-Lab recipe published measured dual-HCA numbers on **two independent 2x GB10 TP=2 pairs**
(direct QSFP), isolating the HCA variable with MTU held constant:

```
Pair 1:  1311 tok/s  MTU 1500, one HCA
         1343 tok/s  MTU 9000, one HCA
         1410 tok/s  MTU 9000, both HCAs     <- +5% from the second controller alone
```

They also report, matching what we saw: no material decode change, "consistent with latency-bound
all-reduces", and explicitly — "Do not apply the nccl-tests busbw delta (+64% here) directly to
end-to-end throughput: the all-reduce is only a slice of prefill step time."

So the second rail does carry a real prefill gain, and my GID-3 attribution was most likely an
artifact of a bad GID-0 arm rather than evidence for GID 3. Our GID-0 run was n=2 with a 0.68 s
internal spread; theirs is a properly isolated comparison across two pairs. Theirs wins.

Two things from their work worth folding in here regardless of this PR:

1. They independently document the same default this PR is about — `NCCL_IB_MERGE_NICS` already
   defaults to 1, it *permits* merging rather than selecting HCAs, and **the meaningful override
   is `0`**. That is exactly the argument for the pass-through: without it you cannot deliver the
   opt-out.
2. They resolve the RoCE v2 GID **per node from sysfs** (`NCCL_IB_GID_AUTO=1`) rather than pinning
   a literal, because "index differs per node/HCA and drifts after reboot; a single shared literal
   on both ranks can wedge NCCL at init." We pin `NCCL_IB_GID_INDEX=3`; it is correct on all four
   of our HCAs today (verified in sysfs), but their point about post-reboot drift is well taken and
   we have added a drift check rather than assume it stays valid.

Net for this PR: the fix is unchanged and still needed. What changes is my claim about *what the
second rail buys* — it buys more than I credited it with in the previous comment.

tonyd2wild:
Verified against the repo and merging. Your diagnosis is exactly right:

```
.env.dspark.example:30      NCCL_IB_MERGE_NICS=1
DEFAULT-CONFIG.md:174       "98 Gb/s single vs 161 Gb/s with both HCAs and NCCL_IB_MERGE_NICS=1"
docker-compose.dspark.yml   0 occurrences
```

So the flag credited with the headline fabric number has never once reached the engine. `set -a; source .env.dspark; set +a` puts it in the launcher's shell and it dies there, because compose without `env_file:` only forwards names it explicitly lists.

**The framing in your PR is the valuable part** — *"a variable you did not configure is indistinguishable from one that silently did not arrive."* That's the same class as `DSPARK_MODEL=` being ignored on the command line, and it's a class worth designing against rather than fixing case by case. Anyone who read the 161 Gb/s line and assumed their fabric was tuned has been running at 98.

I hit the mirror image of this on the GLM TP4 fleet today and it's worth recording next to yours: there, `NCCL_IB_GID_INDEX` *was* passed through correctly but **pinned**, and a pin goes stale when a link bounces or an address is re-added — the GID moves to the next free slot and the pin points at a hole. Different mechanism, identical symptom class: a fabric setting that looks configured and isn't in effect. That fleet now looks the index up from sysfs at launch and prints what it chose.

Both cases argue for the same rule: **anything fabric-related should be echoed at launch from inside the container**, not trusted from the config file. Worth a follow-up so operators can see what the engine actually got rather than what the env file said.

Issue #38, opened 2026-08-21
Dual-HCA field report: works on Anemll 0.1.1 vLLM stack — three pitfalls that masked it (netplan-volatile IPs, MTU reset after FW reboot, GID index drift → don't pin NCCL_IB_GID_INDEX)


Lollorosso2:
Following the dual-HCA QSFP doc (#35) on 2x GB10 (GIGABYTE AI TOP ATOM, DGX OS, direct QSFP pair) serving DeepSeek-V4-Flash-0731 TP=2 on the Anemll dspark-vllm-gx10:0.1.1 image (MiaAI-Lab recipe): **merged NICs work end-to-end**, but three stacked pitfalls each produced silent TP-handshake hangs that looked like 'merge is broken'. Posting so others don't burn a night on the same stack.

**1. nmcli persistence is an illusion on DGX OS.** netplan owns the config; NM profiles live in /run and any `nmcli con mod` evaporates on reboot. Persist the second NIC via a netplan drop-in (addresses + mtu 9000), not nmcli.

**2. A firmware reboot (fwupdmgr capsule) reset the second NIC to MTU 1500** while the peer stayed at 9000 → RDMA `IBV_WC_RETRY_EXC_ERR(12)` on large transfers, plus NCCL falling back to fe80 link-local GIDs once the IPv4 GID vanished with the address.

**3. GID table indexes drift.** With both HCAs up, the RoCEv2/IPv4 GID moved from index 3 to index 4 on `roceP2p1s0f1` within 30 minutes (interface events reshuffle the table; index 3 became a hole). A pinned `NCCL_IB_GID_INDEX=3` (per the doc) then fails only on the second HCA: `ibv_modify_qp failed with 61` + silent hang after model load. **Fix: leave NCCL_IB_GID_INDEX entirely unset** (compose: map it to `null`, an empty string parses as 0 = fe80). Modern NCCL selects the RoCEv2/IPv4 GID per device and is immune to the drift. With the pin removed, the merged boot went from silent-hang/36-min to ~5 min, zero NCCL warnings.

[...]

PR #40, opened 2026-08-21
docs: dual-HCA — don't pin NCCL_IB_GID_INDEX (GID drift, #38) + netplan/MTU/JIT-retune pitfalls

Capicua25x:
Folds @Lollorosso2's #38 field report back into the dual-HCA doc, plus validation from the reference pair.

**The correction:** the doc recommended `NCCL_IB_GID_INDEX=3` (verified-at-write against `gid_attrs`). #38 showed GID table indexes **drift** when interface events reshuffle the table (3→4 within 30 min observed) — a pinned index then fails on ONE HCA only (`ibv_modify_qp` err 61) and hangs the TP handshake silently after model load. Modern NCCL selects the RoCEv2/IPv4 GID per device; the doc now says **leave it unset**, with the compose caveat that matters: map the variable to `null` — an empty string parses as 0 = the fe80 link-local GID, which is worse than the pin.

**Also folded from the report:** persist the second NIC via a netplan drop-in on DGX OS (nmcli profiles live in /run), re-verify MTU 9000 both ends after firmware capsule reboots (`IBV_WC_RETRY_EXC_ERR(12)` signature), and expect ~36 min of JIT retuning on the first boot after any NCCL topology change — a 20-minute watchdog will 'confirm' a hang that isn't one.

**Validated on the reference pair today (2026-08-21):** the same latent pin was found live in our `.env.dspark` (both nodes), removed with the compose-null pattern, and the unpinned boot came up clean — start script exit 0, `NCCL_IB_GID_INDEX` absent from container env, zero NCCL warnings, TP generation verified.

Files: `.env.dspark.example` (pin retired with rationale), `DEFAULT-CONFIG.md` (env line + dual-HCA note rewritten with all three pitfalls, credit in-text).

https://claude.ai/code/session_01JYBRsZkZKH3aV1tDwkmS8H

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -28,9 +28,15 @@ HEADLESS=
 NCCL_IB_HCA=rocepXsYfZ,rocePWpXsYfZ
 NCCL_SOCKET_IFNAME=enpXsYfZnpN,enPWpXsYfZnpN
 NCCL_IB_MERGE_NICS=1
-# GID 3 = RoCEv2. Verify per port before serving:
-#   cat /sys/class/infiniband/<hca>/ports/1/gid_attrs/types/3   -> "RoCE v2"
-NCCL_IB_GID_INDEX=3
+# Do NOT pin NCCL_IB_GID_INDEX (changed 2026-08-21, issue #38 field report):
+# GID table indexes DRIFT when interface events reshuffle the table (observed
+# 3->4 within 30 minutes with both HCAs up). A pinned index that was correct at
+# write time later fails on ONE HCA only ("ibv_modify_qp failed with 61") and
+# hangs the TP handshake silently after model load. Modern NCCL selects the
+# RoCEv2/IPv4 GID per device on its own. If your compose maps this variable,
+# map it to null — an EMPTY STRING parses as 0, which selects the fe80
+# link-local GID and is worse than the pin.
+# NCCL_IB_GID_INDEX=3   # <- do not re-enable; kept for archaeology
 NCCL_CROSS_NIC=1
 
 # Caches / model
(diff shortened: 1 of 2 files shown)


tonyd2wild:
Merging — the pin is genuinely fragile and retiring it is right. One important caveat to record alongside it, because a sibling fleet hit the **opposite** failure from the same root cause.

**Confirming the drift risk is live here.** Probing the reference pair's dual-HCA setup just now:

```
gx10-2f59  rocep1s0f0    -> RoCEv2/IPv4 at index 3
gx10-2f59  roceP2p1s0f0  -> RoCEv2/IPv4 at index 3
spark-3977 rocep1s0f0    -> RoCEv2/IPv4 at index 3
spark-3977 roceP2p1s0f0  -> RoCEv2/IPv4 at index 3
```

Uniform at 3 across both HCAs on both nodes — so the pin happens to be correct *right now*, which is exactly the trap @Lollorosso2 describes. A config that is correct until an interface event silently becomes a pin into a hole, and then fails on one HCA only.

**The caveat: `unset` is not universally safe.** On the GLM-5.3 4-node fleet ([TP4 #6](https://github.com/tonyd2wild/GLM-5.3-Flash-NVFP4-1M-KV-4x-DGX-Spark/pull/6), @tsw2k, four GX10 through an Arista 7060CX-32S at 100G), all three states were characterised:

| GID setting | failure |
|---|---|
| wrong explicit index | `ibv_modify_qp failed ... local GID ::` — clear |
| **stale pin** | `NCCL error: unhandled system error` on one rank; head says only "WorkerProc initialization failed" |
| **unset (auto-select)** | **collective hangs, no log output at all** |

So on that fabric, unset is the *worst* of the three. Your pair validates unset cleanly; theirs hangs on it. Both reports are real — plausibly dual-HCA-2-node vs switched-4-node, or an NCCL version difference in the per-device selection path.

**The form that survives both** is to look the index up from sysfs at launch and pass it explicitly, defaulting to the old value if nothing matches. That's what the GLM TP4 launcher now does — six lines, prints what it picked, no drift because it's resolved fresh every boot, and no reliance on auto-select. It's strictly more robust than either pinning a constant or leaving it unset.

I'd suggest that as a follow-up here rather than blocking this PR, since your change is correct for this fabric and removes a live latent pin today.

**The other three items are excellent and I'd have lost a day to each:**
- netplan drop-in because nmcli profiles live in `/run` and evaporate — non-obvious and permanent-looking until reboot
- re-verify MTU 9000 after firmware capsule reboots, with the `IBV_WC_RETRY_EXC_ERR(12)` signature to recognise it by
- **~36 min of JIT retuning after an NCCL topology change, and a 20-minute watchdog will "confirm" a hang that isn't one.** That one is worth its own line in the troubleshooting section — a monitor that manufactures the failure it's watching for is a particularly expensive kind of wrong.

Also relevant to @tsw2k's diagnostic on the other fleet: the tell for a *real* NCCL hang is **96% GPU utilisation at ~18 W**. That's a spin-wait, not compute — a GB10 genuinely working draws far more. Pairs well with your 36-minute JIT note, since it distinguishes "slow" from "wedged" in five seconds.

PR #42, opened 2026-08-29
Patch 5: also arm the stop-in-reasoning guard on output-side think tags

Mcray4:
The guard armed only when the prompt's last token was the reasoning start marker. Templates that ignore the thinking chat_template_kwarg (prompt token ids identical either way) never put the tag in the prompt — the model opens <think> itself as its first output token — so the guard never armed and stops still truncated mid-reasoning (measured: stop ["17 * 23"] on a math prompt, 90 completion tokens, empty content AND empty reasoning_content, on a DeepSeek-V4-Flash-0731 derivative).

New: with stops present and reasoning markers configured but no prompt-side tag, the guard arms PENDING; the first visible output chunk resolves it. Opening with the start marker (leading whitespace tolerated, marker-prefix stays pending) promotes to the full guard; anything else evaluates the deferred span with normal stop semantics so a stop in the very first characters still fires. Requests without stops untouched.

Validated in the serving container on a 2x DGX Spark TP=2 pair: four branches driven directly against the patched class (reasoning-dormant / deferred-span fires / prompt-side unchanged / leading whitespace), plus live end-to-end content-stop checks; single-stream peak unchanged. The patch file's WHY/WHAT/VALIDATION sections carry the details. Relates to issue #18.

### DIFF
--- a/patches/0005-suppress-stops-in-reasoning.patch
+++ b/patches/0005-suppress-stops-in-reasoning.patch
@@ -1,8 +1,9 @@
 From: Capicua25x
+Extended-by: Mcray4 (output-side arming)
 Subject: [PATCH 0005] detokenizer: don't evaluate client stop strings inside the
- reasoning segment
+ reasoning segment — prompt-side AND output-side think tags
 
-Fixes null/empty content on think-in-prompt models when the client sends stop
+Fixes null/empty content on thinking models when the client sends stop
 strings. See issue #18.
 
 WHY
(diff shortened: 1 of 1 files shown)


tonyd2wild:
Merging. This closes a real hole in the guard's arming condition, and the failure mode you describe is the nastiest kind — silent and total.

The gap in one line: the guard armed on **prompt-side** evidence (last token == reasoning start marker), but templates that ignore the `thinking` chat_template_kwarg produce identical prompt token ids either way, so the tag never appears in the prompt — the model opens `<think>` itself as its first *output* token. The guard never armed, and stops truncated mid-reasoning. Your measurement is the part that makes it undeniable: **90 completion tokens, empty `content` AND empty `reasoning_content`.** A response that consumed budget and returned nothing at all, with no error.

The PENDING-then-resolve-on-first-chunk design is the right shape. Arming on prompt evidence was inferring intent from the wrong side of the boundary; the first visible output chunk is where the answer actually is. I like that the deferred span still evaluates with normal stop semantics, so a stop in the very first characters isn't swallowed by the fix — that's the obvious way to over-correct and you avoided it.

Four-branch validation against the patched class plus live end-to-end, with single-stream peak unchanged, is more than enough. Marker-prefix staying pending is a nice detail — partial `<th` arriving in one chunk would otherwise resolve wrong.

Relates to #18, and I'll fold this into the assessment there.

PR #43, opened 2026-08-31
Native vision: deploy DeepSeek-V4-Flash-Vision-Exp on 2x DGX Spark (day-0)

tonyd2wild:
DeepSeek released [**DeepSeek-V4-Flash-Vision-Exp**](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash-Vision-Exp) today. This deploys it on 2× DGX Spark at TP2 **with working native image input** and DSpark speculative decoding intact.

```
"Based on the image, the two colors are red and blue.
 Red is on the left side. Blue is on the right side."
```

Not a sidecar and not a VLM proxy — the model's own 32-block ViT and aligner running inside vLLM.

## Why a port was needed

vLLM's `DeepseekV4ForCausalLM` is the **text-only** class. The vision checkpoint reports the *same* `architectures` string while carrying 316 extra tensors vLLM has nowhere to put, so it doesn't load at all:

```
ValueError: There is no module or parameter named 'aligner' in DeepseekV4ForCausalLM
```

DeepSeek shipped only a reference implementation — its own `inference/README.md` calls it "a readable reference implementation rather than a production serving engine." There was nothing to configure.

## What's actually new in this checkpoint

Diffing configs, the language model is identical to 0731 except:

| | 0731 | Vision-Exp |
|---|---|---|
| vision keys | — | 10 (32-block ViT, dim 1024, patch 14, 2D RoPE) |
| `num_nextn_predict_layers` | 1 | **3** |
| `rms_norm_eps` | 1e-6 | 1e-20 |
| `ffn.gate.*` | `weight`, `tid2eid` | + **`bias`**, **`bias_vl`** |

That last row is a **modality-specific MoE routing bias** — image tokens route to different experts than text tokens — present on all 43 layers including the 3 hash-routing layers 0731 left unbiased. Undocumented in any paper or model card.

## Twelve blockers

Full list with errors in [`vision-exp/README.md`](vision-exp/README.md). The three most likely to catch anyone else:

- **vLLM answers `is_multimodal_model` from a static architecture-name table**, never inspecting the class. `SupportsMultiModal` in the MRO does nothing on its own — needs a registry alias plus `--hf-overrides`.
- **That inspection is cached on disk** in `$VLLM_CACHE_ROOT/modelinfos/`, keyed by module+class. Both names resolve to the same class, so the alias silently reused the stale pre-patch "text-only" entry. Clear `modelinfos/` after changing model interfaces.
- **DeepSeek-V4's first `num_hash_layers` MoE layers route by token id**, but vLLM's multimodal path passes `inputs_embeds` with `input_ids=None` → `DeepSeek V4 hash MoE routing requires input_ids`. Fixed with `requires_raw_input_tokens = True`.

## Measured (2× GB10, TP2, temperature 0)

| | |
|---|---|
| Text, count-to-100 | **55.9 tok/s** median · 58.0 peak |
| DSpark acceptance | 0.628 · **4.14** mean accepted length (k=5, max 6) |
| Vision, 336×336 + 26-token answer | **1.03 s** end to end |

Correctness verified on two discriminating tests (different colors, different split axis, different image sizes) — both exactly right, and image-block token counts match the reference math.

## Not parity — two documented deviations

1. **Bidirectional attention within image spans is not implemented.** The reference widens the sparse-attention window so image tokens attend bidirectionally; here they use the causal sparse pattern. Closest measured analogue (vLLM #40106, Gemma-4) shows KL 0.03–0.09 concentrated at image positions. Expect degradation on OCR/chart/document work, and note it still passes smoke tests.
2. **`bias_vl` is loaded but not applied** — image tokens currently route through the text gate bias.

Neither affects text-only requests. Every patch is guarded on `vision_n_layers`, so **0731 is unchanged**.

## Prior art

vLLM [#54561](https://github.com/vllm-project/vllm/issues/54561) and draft PR [#54566](https://github.com/vllm-project/vllm/pull/54566) went up the same day with a parallel implementation on 2× RTX PRO 6000. What's new here is **GB10 / DGX Spark**, where the [NVIDIA forum position that day](https://forums.developer.nvidia.com/t/deepseek-v4-flash-vision-exp-is-released-as-open-weights/381911) was that the native vLLM vision processor would not work with this model, and the previous answer for vision on Spark was a sidecar VLM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,7 @@ reference implementation, explicitly "rather than a production serving engine."
 
 Running the repo's existing **validated agent-serving profile** — `MAX_MODEL_LEN=1500000`,
 `MAX_NUM_SEQS=12`, `GPU_MEMORY_UTILIZATION=0.85`, `MTP_NUM_TOKENS=3` — with the vision port
-on top. **Use `k=3`, not the 0731 recipe's `k=5`:** this release moved
-`num_nextn_predict_layers` from 1 to 3, and carrying k=5 across measurably underperforms.
+on top. **Use `k=5`, same as the 0731 recipe, with Patch 4 mounted.** An earlier version of this branch said k=3: that A/B was measured without the Patch 4 `spec-dspark.py` mount, which silently halves the draft's acceptance (see `vision-exp/README.md`, [#48](../../issues/48)). With Patch 4 present, k=5 wins count-to-300 by a third and code is neutral. This release moved `num_nextn_predict_layers` from 1 to 3, which is why the drafter deserves a second look, but not at k=3.
 
 **→ [Full guide, the twelve blockers, and the port: `vision-exp/`](vision-exp/README.md)**
 
@@ -48,7 +47,7 @@ Text-only `0731` is unchanged and still fully supported — every patch is guard
 >
 > **Covers all three checkpoints:**
 > - **`deepseek-ai/DeepSeek-V4-Flash-Vision-Exp`** (2026-08-31, multimodal) — **native vision,
->   55.9 tok/s text.** See [`vision-exp/`](vision-exp/README.md).
+>   55.9 tok/s text, measured before the Patch 4 mount was added to the vision launcher; with Patch 4, #48 measures 85.5 count / 49.9 code / 29.5 prose.** See [`vision-exp/`](vision-exp/README.md).
 > - **`deepseek-ai/DeepSeek-V4-Flash-0731`** (official release) — **78 tok/s peak, ~55 typical.**
 >   Requires [Patch 4](patches/0004-dspark-shared-expert-gate-up-proj.patch); without it you get
 >   roughly half speed at unchanged output quality. **Start here →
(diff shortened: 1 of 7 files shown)


BPAMLUX:
One data point before this merges: the launcher in this branch (f6bf1d8) still lacks the Patch-4 spec-dspark.py mount that PR #44 fixed on main — merging as-is would reintroduce that gap, and the README's k=3-over-k=5 numbers appear to have been measured in that state. Repro + k=3/k=5 A/B numbers on a second 2x DGX Spark: https://github.com/tonyd2wild/DeepSeek-v4-Flash-Vision-Exp-DSpark-1M-NVFP4-KV-2x-DGX-Spark/issues/48

tonyd2wild:
You're right, thanks for the repro. Pushed 30f98a2 (merged main in as 148811c): the vision launcher now mounts the Patch 4 spec-dspark.py right after the Patch 3 mount with a preflight that refuses to start without it, the default is k=5 like main, and the READMEs mark the k=3 A/B as measured without Patch 4 and point at your #48 numbers. I'll re-measure on this rig once the boxes are free and update the table.

PR #45, opened 2026-08-31
Add self-contained sparkrun recipe for DeepSeek-V4-Flash-Vision-Exp

Marker689:
## Summary

Adds a self-contained **sparkrun recipe for `deepseek-ai/DeepSeek-V4-Flash-Vision-Exp`** (`sparkrun/deepseek-v4-flash-vision-exp-dspark-nvfp4-1m-vllm.yaml`) plus a short note in `sparkrun/README.md`.

The repo already ships a sparkrun recipe for the **text** 0731 model, but the vision build is only run via the manual `vision-exp/ds4-vision-tp2.sh` compose/bind-mount launcher. This PR provides the sparkrun equivalent, so the vision port can be deployed the same way (one command, no manual node staging).

## What the recipe does

It is the text recipe plus the vision port, made fully reproducible via `pre_exec` — nothing is staged by hand:

- Fetches the DSpark overlay + nvfp4 stage A/B/C patches from this repo (pinned to `c6673bc`) and reproduces the `dspark-nvfp4-stage-c` runtime in-container (Patch 3 + Patch 4 verified), exactly like the text recipe.
- Fetches the **vision port** (`vision-exp/port/`, commit `d39f94a`) from this repo's `vision-exp-default` branch and injects it: patches `vllm/models/deepseek_v4/nvidia/model.py` and `vllm/model_executor/models/registry.py`, installs `ds4v_vision.py` / `ds4v_mm.py`. **This is why the port is on a feature branch, not main — the recipe points at it explicitly.**
- Clears the `modelinfos/` inspection cache (vLLM caches `is_multimodal_model` per module+class; a stale entry makes it reject images with "is not a multimodal model").
- Patches torch 2.11's `AOTAutogradCache` to use a distinct `aotautograd_functorch` subdir — with shared/raced JIT cache dirs it hits a `FileExistsError` in `os.makedirs` during cold compile (two torch cache layers write file/dir into the same namespace).
- Runs a real import check on the patched model module before the long boot, so an API mismatch fails fast instead of at the end of a 6-minute boot.

## Key decisions (and why)

- **Model**: official `deepseek-ai/DeepSeek-V4-Flash-Vision-Exp`. Swap in any abliterated Vision-Exp derivative by changing the `model:` field — the port is the same.
- **`k` (num_speculative_tokens): 5** for parity with the text recipe / `VISION-EXP-DEFAULT-CONFIG.md`. Note in the header and the `speculative_config` comment: the `vision-exp-default` branch measured **k=3** better (accept 0.830 / 54.6 tok/s vs 0.542 / 52.1 tok/s at k=5 on count-to-300) because this checkpoint has `num_nextn_predict_layers=3` — a valid drop-in if you prioritize measured peak acceptance.
- **`--hf-overrides`** selects the multimodal registry alias (`DeepseekV4VForConditionalGeneration`). Required — vLLM answers `is_multimodal_model` from a static arch-name table, so without it images are rejected.
- **JIT/compile caches forced node-local** (`/cache/runtime`). The text recipe documents overriding the seven cache vars onto a node-local path when the HF cache is shared over NFS; here we set them node-local up front, because a shared HF cache between the two TP ranks races `torch.compile` and corrupts the vision encoder's replicated compile.
- **`thinking:false`** default (the checkpoint ships no Jinja chat template). Header comment documents how to enable reasoning.
- **No `--max-cudagraph-capture-size`** (vLLM derives `max_num_seqs x (k+1)`; a stale literal truncates capture).

Profile: `:8888`, `max_num_seqs 12`, `max_num_batched_tokens 8192`, `gpu_memory_utilization 0.85`, `max_model_len 1048576` — the vision port's verified-live lane.

## Validation

Validated on a real 2x DGX Spark pair: weights load 48/48, memory profiling and torch.compile pass, `Application startup complete`, KV pool ~2.57M tokens; a text request and a split-colour image request both answered correctly (the image path identified red-left / blue-right).

## Known port deviations (not introduced by this PR)

The vision port itself (from the `vision-exp-default` branch) has two quality-affecting, non-crashing deviations documented in `vision-exp/README.md`: image spans use causal sparse attention (no bidirectional window) and `bias_vl` is loaded but not applied. Text-only requests are byte-identical to 0731.

### DIFF
--- a/sparkrun/README.md
+++ b/sparkrun/README.md
@@ -14,7 +14,35 @@ Everything here was deployed and measured on a real 2x DGX Spark pair on
 [main README](../README.md) is the reference — this file is just how to run
 it.
 
-## Quick start
+## Vision-Exp (native image input)
+
+The companion recipe "deepseek-v4-flash-vision-exp-dspark-nvfp4-1m-vllm.yaml" serves the
+experimental VISION build "deepseek-ai/DeepSeek-V4-Flash-Vision-Exp" on the same runtime.
+It is the text recipe plus the vision port (a 32-block ViT + aligner inside vLLM), fetched
+and injected automatically by pre_exec from this repo's "vision-exp-default" branch
+(commit "d39f94a") — it is not in "main". Nothing has to be staged by hand on the nodes.
+
+```bash
+sparkrun run ./deepseek-v4-flash-vision-exp-dspark-nvfp4-1m-vllm.yaml
+```
+
+```bash
+curl -fsS http://<head-ip>:8888/v1/models   # served_model_name deepseek-v4-flash-vision-exp
+# text request; then an image request (image_url base64) to exercise the vision path.
+```
+
+Config notes carried in the recipe header: this checkpoint ships NO Jinja chat template
+(encoding/ scripts only) so "thinking:false" is default; "--hf-overrides" selects the
+multimodal registry alias (without it vLLM answers "is not a multimodal model"); the
+compile/JIT caches are forced node-local ("/cache/runtime") because sharing the HF cache
+over NFS between the two ranks races torch.compile (see the header comment and the text
+recipe's NFS warning). "k" defaults to 5 for parity with the text recipe; the
+"vision-exp-default" branch measured k=3 better (accept 0.830 vs 0.542) since this
+checkpoint has "num_nextn_predict_layers=3" — swap if you prefer measured peak acceptance.
+
+Known port deviations (image quality only, not crashes) and the full write-up are in
+vision-exp/README.md in this repo.
+
 
 Once, on your head node (~200 GB free disk per node):
 
--- a/sparkrun/deepseek-v4-flash-vision-exp-dspark-nvfp4-1m-vllm.yaml
+++ b/sparkrun/deepseek-v4-flash-vision-exp-dspark-nvfp4-1m-vllm.yaml
@@ -1,0 +1,337 @@
+# DeepSeek V4 Flash Vision-Exp (DSpark) — native image input, 1M ctx, NVFP4 KV — 2x DGX Spark
+# sparkrun recipe for the experimental VISION build DeepSeek-V4-Flash-Vision-Exp.
+# It is the text 0731 recipe + the vision port, fully self-contained: pre_exec fetches
+# the vision port from this repo's `vision-exp-default` branch (it is NOT in main — see
+# VISION-EXP-DEFAULT-CONFIG.md) and injects it into the runtime, so nothing must be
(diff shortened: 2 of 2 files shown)


tonyd2wild:
Merging — this is the deployment path the vision port needed, and the `pre_exec` reproduction is exactly right: nothing staged by hand, everything pinned.

Particularly good calls:

- **Clearing `modelinfos/`.** vLLM caches `is_multimodal_model` per module+class, and the alias resolves to the same class, so a stale "text-only" verdict survives every fix you make. I lost an afternoon to precisely that. Baking the clear into the recipe means nobody else has to.
- **Real import check before the long boot.** Failing an API mismatch in seconds instead of six minutes in is the difference between a tight loop and a wasted afternoon.
- **The torch 2.11 `AOTAutogradCache` subdir fix.** Two cache layers writing file-and-dir into the same namespace is not something you'd find by reading; that's a debugged-it artifact.
- **Node-local JIT caches up front**, rather than as a documented override. A shared HF cache racing `torch.compile` across TP ranks corrupting the *vision encoder's replicated compile* is a failure mode specific to this port, and you got ahead of it.

**Two things to update, neither blocking:**

**1. The k=3 note cites my numbers, and my numbers were wrong.** The header references the `vision-exp-default` measurement (accept 0.830 / 54.6 tok/s at k=3 vs 0.542 / 52.1 at k=5). That A/B ran on a launcher **missing the Patch 4 `spec-dspark.py` mount** — see #48, which reproduced it properly on a second pair. With Patch 4 present, k=5 wins count-to-N by **+33%**, code is neutral (49.9 vs 49.5) and only prose modestly favours k=3 (32.1 vs 29.5). Their diagnosis of my result is exact:

> *the branch README's 0.542-accept-at-k=5 matches the unpatched-loader acceptance signature rather than a property of the vision drafter*

So your choice of **k=5 for parity is correct on the merits**, not just for consistency — but the justification in the header should point at #48 rather than my superseded A/B, or it propagates a number that's an artifact of a missing mount.

**2. The port is on `main` now.** The recipe pins `vision-exp/port/` at `d39f94a` on the `vision-exp-default` branch, with a note explaining that's why it lives on a branch. `vision-exp/port/` is on `main` today (`patch_vision.py`, `patch_registry.py`, `ds4v_vision.py`, `ds4v_mm.py`), and #49 added `vision-exp/build-ds4v-files.sh` which does the generate-and-verify step in one command. Repointing at main would drop the branch dependency and pick up the marker verification.

Merging as-is since both are documentation-level and the recipe works; happy to push the two updates myself if you'd rather not.

PR #54, opened 2026-09-06
Patch 6: keep the prefix cache alive on long conversations (protection cap + SWA page recycling)

portavales:
Long-context agent sessions re-prefilled ~400K tokens (235 s) on the first call of most turns, although an identical resend hit 100%. Both causes are in `vllm/v1/core/single_type_kv_cache_manager.py`; this PR adds a single-file overlay (91 changed lines), env-gated, with the old behaviour one variable away.

## Root cause

1. **Unbounded prompt-block protection.** `MLAAttentionManager.cache_blocks` (opted in for `model_version == "deepseek_v4"`) and `SlidingWindowMLAManager.cache_blocks` call `_protect_prompt_blocks`, which `touch()`es prompt pages — an extra `ref_cnt` that outlives the request. The only cap, `2 * max_model_len / block_size`, is 8,192–524,288 pages per group at `max_model_len=1M`, i.e. larger than the whole pool (12,621 pages). The protected set therefore grows ~30–60 pages on every request (+0.45–0.79 pp of the pool), never shrinks while idle, and is released only FIFO under allocation pressure — evicting the oldest conversation's prefix first.
2. **Sliding-window churn through the shared LRU.** DSv4's compressor/indexer groups use pages of a few tokens. `remove_skipped_blocks` frees each skipped page to the global free queue and the next `allocate_new_blocks` pops fresh pages from the queue head. One 354K-token prefill cycles the whole pool several times and evicts every other request's cached pages — with any amount of free space. (1) was the shield against (2).

Effect for an agent: the first call of a user turn (which cannot hit at the previous prompt boundary once the prior round's reasoning is stripped) re-prefills ~400K tokens instead of 1–4 s, and any large intervening request destroys the conversation's prefix.

## Fix

- Cap the protected set to a fraction of the pool, shared across the KV groups: `VLLM_PROTECTED_PROMPT_BLOCKS_FRACTION` (default 0.30; `-1` restores the unbounded behaviour).
- Sliding-window groups keep the pages they skip in a per-request recycle list (`ref_cnt` kept, hash evicted) and reuse them for that request's next allocations; leftovers are freed in `free()`. Shared or protected pages (`ref_cnt != 1`) are freed as before. `VLLM_SWA_RECYCLE_SKIPPED_BLOCKS=0` disables it. Admission accounting is unchanged (conservative: it ignores recyclable pages).

Both launchers pass `VLLM_PROTECTED_PROMPT_BLOCKS_FRACTION=${PROTECTED_FRACTION:-0.30}` and `VLLM_SWA_RECYCLE_SKIPPED_BLOCKS=${SWA_RECYCLE:-1}`, mount `/var/tmp/patch6-single_type_kv_cache_manager.py` read-only like Patches 3/4, and check it in the preflight.

## Validation (2× DGX Spark, TP=2, DeepSeek-V4-Flash-Vision-Exp, DSpark k=5)

```
                                        stock image          cap only             cap + recycle (this PR)
354K prompt, identical resend           100%  1.3 s          100%  1.3 s          100%  1.2 s
second 354K tree, then first again      0%    236 s (cold)   0%    229 s (cold)   100%  1.2 s
second tree again                       0%    235 s (cold)   0%    230 s (cold)   100%  1.3 s
third tree, then first / second         —                    —                    100% / 100%
first tree + 300-token decode, resend   —                    —                    100% (6.8 s decode)
idle KV usage per small request         +0.47 pp             +0.46 pp             +0.48 pp until the cap engages
idle KV usage after 80 small requests   ~37% (extrapolated)  15.2%                capped
greedy output vs stock                  reference            identical            identical
1200-token prose generation, no think   —                    —                    coherent, 35.6 tok/s
reasoning_effort=max                    —                    —                    reasoning + answer returned
```

Bisection from a clean boot (`ratchet-*.log`): removing `--async-scheduling` changes nothing (+0.45 pp/request); removing DSpark reduces the residue to +0.14 pp but does not remove it — DSpark amplifies the problem, the cause is the KV manager.

Analysis and patch by Claude (Fable 5.1) via Claude Code; run and validated on my cluster.

## Files

- `recipe/overlay/vllm/v1/core/single_type_kv_cache_manager.py` — overlay file (based on the probe-c image's module)
- `patches/0006-kv-cache-prompt-protection-cap-and-swa-recycle.patch` — the 91-line unified diff
- `docs/PATCH6-KV-CACHE-PREFIX-EVICTION.md`, entry in `docs/PATCHES.md`, README staging line
- `docs/patch6-validation/` — raw logs behind every number above
- `launchers/ds4-vision-tp2.sh`, `launchers/ds4-vision-tp4.sh` — preflight guard, read-only mount, two env knobs
- `CURRENT.md` (preflight bullet + knobs, launcher hashes refreshed), `.env.dspark.example` (`PROTECTED_FRACTION`, `SWA_RECYCLE`)

---

**Which line of CURRENT.md does this change?**
The preflight bullet: "Patch 3 (`patch3-scheduler.py`), Patch 4 (`spec-dspark.py`) **and** the four vision port files … The launcher checks all six and exits if any is missing." — now seven staged files (Patch 6 added) plus a new bullet documenting the two env knobs. Launcher hashes refreshed with `bash tools/check-current.sh --write`.

**What is the new value, and what number proves it?**
New: `VLLM_PROTECTED_PROMPT_BLOCKS_FRACTION=0.30` and `VLLM_SWA_RECYCLE_SKIPPED_BLOCKS=1` (both launchers) with the Patch 6 mount. Before → after: a 354,481-token prompt re-sent after another 354K-token prefill hits 0% and re-prefills cold in 236 s → hits 100% in 1.2 s. Measurement: a real agent transcript as the prompt, `temperature=0`, `max_tokens=1`, `reasoning_effort=max`; equivalence check `temperature=0`, `seed=1`, byte-identical to the stock image. Logs: `docs/patch6-validation/validate_patch6.log` (fixed), `replay.log` (stock), `ratchet-{baseline,noasync,nospec}.log` (bisection), `validate_patch5.log` (cap only). No serving flag that affects speed is changed.

**Tested on**
2× DGX Spark GB10, TP=2 over RoCE; `vllm-dspark-runtime` (probe-c / keys-concurrency-p2b family, vLLM `0.21.1rc1.dev339+g1967a5627bc3`) with Patches 3/4 mounted; weights `deepseek-ai/DeepSeek-V4-Flash-Vision-Exp` @ `86f746b36186f0e567729a5c06a8c918caba82a9`; 2026-09-06. The TP4 launcher carries the identical change but was **not run** (I only have two Sparks); the KV manager runs on the head, so the change does not depend on TP size.

****Independent validation:** #55 (second 2× DGX Spark pair, same vLLM line): both halves logged at boot, ~99.8% prefix-cache hit on the next request vs. 0% stock, idle KV ~80% → ~7%.

Checklist**
- [x] I rebased on current `main` (branch base is `d841fdb`, the current `main`)
- [x] `CURRENT.md` updated in this PR if a launcher or a serving flag changed, and `bash tools/check-current.sh --write` was run (`tools/check-current.sh` passes)
- [x] Speed numbers are from real prompts (354K agent transcript for the cache timings, prose essay for tok/s); the 235 s figures are cold prefill, the 1.2 s figures are the warm hits this PR restores; no counting-prompt numbers

### DIFF
--- a/.env.dspark.example
+++ b/.env.dspark.example
@@ -185,3 +185,9 @@ VLLM_SKIP_INIT_MEMORY_CHECK=1
 # nvfp4_ds_mla gives the 1M pool; fp8_ds_mla is the conservative choice
 # when 4-bit KV quality under long agentic context matters more than reach.
 #KV_CACHE_DTYPE=nvfp4_ds_mla
+
+# Patch 6 — prefix-cache retention on long conversations (docs/PATCH6-KV-CACHE-PREFIX-EVICTION.md)
+# Share of the KV pool the DSv4 prompt-block protection may pin, summed over KV groups (-1 = old unbounded behaviour).
+PROTECTED_FRACTION=0.30
+# Sliding-window/compressor groups reuse their own skipped pages instead of churning the global LRU (0 = old behaviour).
+SWA_RECYCLE=1
--- a/CURRENT.md
+++ b/CURRENT.md
@@ -31,9 +31,15 @@ Both launchers read `MODEL_DIR` (default `DeepSeek-V4-Flash-Vision-Exp`); the un
 
 **Preflight, on all four nodes:**
 
-- Patch 3 (`patch3-scheduler.py`), Patch 4 (`spec-dspark.py`) **and** the four vision port files
-  (`ds4v_model.py`, `ds4v_vision.py`, `ds4v_mm.py`, `ds4v_registry.py`) staged at `/var/tmp`.
-  The launcher checks all six and exits if any is missing.
+- Patch 3 (`patch3-scheduler.py`), Patch 4 (`spec-dspark.py`), Patch 6
+  (`patch6-single_type_kv_cache_manager.py`, source `recipe/overlay/vllm/v1/core/single_type_kv_cache_manager.py`)
+  **and** the four vision port files (`ds4v_model.py`, `ds4v_vision.py`, `ds4v_mm.py`, `ds4v_registry.py`)
+  staged at `/var/tmp`. The launcher checks all seven and exits if any is missing.
+- Patch 6 keeps the prefix cache alive on long conversations: both launchers pass
+  `VLLM_PROTECTED_PROMPT_BLOCKS_FRACTION=${PROTECTED_FRACTION:-0.30}` and
+  `VLLM_SWA_RECYCLE_SKIPPED_BLOCKS=${SWA_RECYCLE:-1}`. Measured on TP2: a 354K-token prompt
+  re-sent after another 354K prefill hits 100% in 1.2 s (stock image: 0%, 235 s cold re-prefill);
+  logs in `docs/patch6-validation/`, analysis in `docs/PATCH6-KV-CACHE-PREFIX-EVICTION.md`.
 - Workers mount Bluey's weights export at `/mnt/bluey-models`.
 - **Drop page cache on all four nodes before launch.**
 
@@ -97,5 +103,7 @@ nothing a PR or issue links to changes path — `vision-exp/ds4-vision-tp2.sh` s
 The self-contained `sparkrun/` recipe serves under the id `deepseek-v4-flash-vision-exp`; the launchers here serve `deepseek-v4-flash-dspark`. Clients pointed at :8888 use the launcher's id.
 
 <!-- launcher hashes, maintained by tools/check-current.sh --write -->
-sha256 310e74b5a459f3cab876cd4d5edcbe67416ed74e408dfd965fd592a0bd0b1409  launchers/ds4-vision-tp2.sh
-sha256 3ace8ad8172c5a9c146e7dacc385b16d1e60049f65b7697527ec14eeead35ea4  launchers/ds4-vision-tp4.sh
+
+<!-- launcher hashes, maintained by tools/check-current.sh --write -->
+sha256 63a5f6c3ea32ecf764f76ac32ef7307dd637430301eab0ffc9bf0dbdd5d40531  launchers/ds4-vision-tp2.sh
+sha256 50b33b978f9c96ed318907519ff3f4e60abf15fb762c90345852e3c2cae1b266  launchers/ds4-vision-tp4.sh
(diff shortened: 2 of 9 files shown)


Capicua25x:
Independent validation on a second 2× DGX Spark pair (DeepSeek-V4-Flash-Vision-Exp, TP=2, same vLLM line `0.21.1rc1.dev339+g1967a5627bc3`), deployed as a read-only bind-mount with `VLLM_PROTECTED_PROMPT_BLOCKS_FRACTION=0.30 VLLM_SWA_RECYCLE_SKIPPED_BLOCKS=1`:

- Boot confirms both halves:
  - `Patch 5: protected prompt block cap for KV group 0 = 1037 blocks (pool 17290 blocks, 5 managers, fraction 0.30, legacy cap 8192)`
  - `Patch 6: KV group 3 recycling skipped pages in-request (first request …, 2046 pages)`
- Cold turn: 393,231-token prompt pre-filled in a single clean pass (KV monotonic, no waits), then decode.
- Next request: ~787,894-token prompt with **786,688 cached tokens (~99.8% hit)**. Stock was 0% hit / full ~235 s re-prefill — matching your numbers.
- Idle KV after: **~7.2%** (previously ~80%, released only FIFO under pressure).

Filed the underlying report as #55 (root cause + these numbers) so it is linkable from issues. This fixes a real, reproducible long-conversation failure on the vision port — would be great to see it merged.

portavales:
Thank you for the independent validation — that's the same behaviour on a different pair: both halves logged at boot, one clean cold prefill at ~393K, ~99.8% cache hit on the next request versus 0% on stock, and idle KV back to ~7%. Your larger pool (17,290 blocks at `gpu-memory-utilization 0.85`) also shows the fraction-based cap scaling as intended.

Linking #55 from the PR description so the root cause and both sets of numbers are in one place. The two knobs (`VLLM_PROTECTED_PROMPT_BLOCKS_FRACTION=0.30`, `VLLM_SWA_RECYCLE_SKIPPED_BLOCKS=1`) are already documented in `CURRENT.md` and `.env.dspark.example` in this PR.

